PR #618, opened 2025-07-21
feat: initial parse hooks

tada5hi:


tada5hi:
Closing as superseded: this targets the pre-monorepo v1 layout (root src/, hookable-based fields hooks), and v2 shipped per-parameter validate hooks with context threading (#806/#807) plus condition-returning hooks (#830/#840) that cover the use case. The branch stays available if anything needs salvaging.

PR #728, opened 2026-06-21
fix(core): correct isFilters discrimination and relation extract; add core test baseline

tada5hi:
## Summary

Adds a boundary-test baseline for `@rapiq/core` (it had a single spec/one test) and fixes two real bugs that writing those expected-behavior tests surfaced.

## Fixes

- **`isFilters` never matched** (`parameter/filters/collection/check.ts`) — it called `Array.isArray` on an `IFilters`, which is a plain object, so it always returned `false`. Consequences: `Filters.flatten` silently never hoisted nested same-operator groups, and `ExpressionFiltersParser.parse` double-wrapped every compound (`Filters(AND, [Filters(AND, [...])])`). Now discriminates on the compound-only `flatten` method via `isObject`.
- **`Relations.extract` prefix collision** (`parameter/relations/collection/module.ts`) — the prefix check had no dot boundary, so `extract('item')` wrongly consumed a sibling `items`. Now requires a `root.` boundary.

## Tests

Core suite grows from 1 spec/1 test to 6 specs/61 tests:

- `parameter/fields.spec.ts` — `Fields.execute` include/exclude/explicate + default/allowed fallbacks
- `parameter/filters.spec.ts` — `Filter.accept` per-operator dispatch + `visitFilter` fallback; `Filters.flatten` associativity
- `parameter/relations.spec.ts` — `Relations.extract` mutation + dot boundary
- `schema/schema.spec.ts` — `Schema` normalization + `extendSchemasOptions` propagation
- `schema/registry.spec.ts` — `SchemaRegistry` add/get/getOrFail/drop + `resolve` schemaMapping traversal

Also adds `parser-expression` coverage for `ExpressionFiltersParser.parse` top-level wrapping — the path the `isFilters` fix actually changes, previously untested (all assertions used `parseExact`). The "return an existing compound group without re-wrapping" case would have failed before the fix.

## Verification

- `@rapiq/core`: 61/61 green.
- Rebuilt core; all downstream packages (parser-simple, parser-expression, codec-url-simple, sql, typeorm) green — both fixes are safe across the monorepo.
- Lint clean on all changed source and test files.

## Notes

Both fixes are workarounds for a deeper smell: the filter AST has no discriminant field distinguishing a leaf `IFilter` from a compound `IFilters` (both are `{ operator, value }`), so `isFilters` must duck-type. Resolving that root cause belongs to the upcoming schema-resolution / AST work rather than this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filter type validation to correctly identify compound filter structures.

* **Tests**
  * Added comprehensive test coverage for field selection behavior, filter operations, schema registry, schema definition and normalization, and expression parser functionality.

### DIFF
--- a/packages/core/src/parameter/filters/collection/check.ts
+++ b/packages/core/src/parameter/filters/collection/check.ts
@@ -6,13 +6,22 @@
  */
 
 import type { ICondition } from '../condition';
+import { isObject } from '../../../utils';
 import type { IFilters } from './types';
 
 export function isFilters(
     input: ICondition,
     operator?: string,
 ) : input is IFilters {
-    if (!Array.isArray(input)) return false;
+    // A compound `Filters` node is the only `ICondition` that carries a
+    // `flatten` method; a leaf `Filter` never does. This distinguishes the two
+    // reliably, even when a leaf filter's `value` happens to be an array (e.g. `in`).
+    if (
+        !isObject(input) ||
+        typeof (input as Partial<IFilters>).flatten !== 'function'
+    ) {
+        return false;
+    }
 
     if (operator) {
         return operator === input.operator;
--- a/packages/core/test/unit/parameter/fields.spec.ts
+++ b/packages/core/test/unit/parameter/fields.spec.ts
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2025.
+ * Author Peter Placzek (tada5hi)
+ * For the full copyright and license information,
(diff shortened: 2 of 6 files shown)


tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai review

PR #741, opened 2026-07-06
feat(sql)!: fragment assembly, null semantics, MSSQL LIKE fallback, typed adapter errors

tada5hi:
Completes the `@rapiq/sql` root adapter (roadmap M2, plan 005).

## What

- **`Adapter.build()` → `SqlFragments`** — `{ columns, where, params, orderBy, limit, offset, relations }`. Resolves the plan's open question as *fragments-only*: FROM/JOIN assembly stays with the caller, which knows the table layout; `execute()` remains for backend adapters (e.g. `@rapiq/typeorm`) that mutate a query object. Both `execute()` TODOs and the single-where-clause TODO are gone.
- **Null semantics** (typeorm-extension parity): `eq(field, null)` → `field IS NULL`, `ne` → `IS NOT NULL`, `in` with a null element → `(field IN (...) OR field IS NULL)`, `nin` with null → `(field NOT IN (...) AND field IS NOT NULL)`. Previously null was bound as an ordinary parameter and matched nothing. Applies to `@rapiq/typeorm` too (shared visitor).
- **MSSQL**: `startsWith`/`endsWith`/`contains` (and negations) now work via `LIKE ... ESCAPE '\'` with wildcard escaping (previously threw); `DialectOptions.regexp` became optional and the `regex` operator throws a typed `AdapterError` (`ErrorCode.FEATURE_UNSUPPORTED`) instead of a raw `Error`.
- **Error taxonomy**: core exports `AdapterError` with `operatorUnsupported`/`featureUnsupported` factories plus `ErrorCode.OPERATOR_UNSUPPORTED`/`FEATURE_UNSUPPORTED`; raw `Error` throws in `@rapiq/sql`/`@rapiq/typeorm` migrated.
- Relations adapter no longer duplicates parent path entries; `AdapterOptions` accepts `rootAlias` (forwarded to fields/filters/sort sub-adapters).

## Tests

- `adapter.spec.ts` — full fragment output for all five dialect presets (pg, oracle, mysql, sqlite, mssql), no-alias mode, excluded fields, relation path dedup, `clear()`.
- `interpreters/null.spec.ts` — all null rewrites + continuous placeholder numbering around them.
- `interpreters/regex.spec.ts` — typed MSSQL error, LIKE fallbacks, wildcard escaping.

Docs: `integrations/sql.md` rewritten for the fragment API, null semantics and the MSSQL fallback; `guide/index.md` errors section mentions `AdapterError`. Behavior changes logged in `.agents/migration-notes.md`.

BREAKING CHANGE: null filter values now emit `IS NULL` predicates instead of bound parameters; the mssql preset no longer throws for anchored string operators and throws `AdapterError` (not `Error`) for regex.

## Summary by CodeRabbit

* **New Features**
  * Added clearer SQL fragment building, letting queries expose selected columns, filters, sorting, limits, offsets, and relation paths more explicitly.
  * Added optional root alias support for SQL query generation.

* **Bug Fixes**
  * Corrected `null` handling so equality, inequality, and list filters use proper SQL null predicates.
  * Improved string matching fallbacks for databases without regex support, including escaped `LIKE` matching.
  * Replaced generic unsupported-operator errors with typed, more consistent adapter errors.

* **Documentation**
  * Updated SQL and error docs to reflect the new query behavior and dialect support.


tada5hi:
@coderabbitai pause

PR #742, opened 2026-07-06
fix: compound wrapping, anchored regex patterns, empty IN lists, sqlite preset

tada5hi:
Fixes the four confirmed correctness findings from the #741 review audit, plus the copy-paste cleanups that let them regress silently.

## Correctness fixes

- **Compound wrapping** (`@rapiq/sql`): `FiltersBaseAdapter.merge()` decided parenthesization via `sql[0] !== '('`, so a nested compound whose *first* condition was itself parenthesized — e.g. the null-rewrite fragment `("realm_id" in($2) or "realm_id" is null)` — was merged unwrapped, producing wrong AND/OR precedence: `and(eq, or(in-with-null, eq))` rendered as `(name AND in-group) OR age` instead of `name AND (in-group OR age)`. Wrapping is now decided by condition count (single conditions are atomic terms or self-wrapped compounds). A top-level multi-condition OR group is now always wrapped too — the old expectation in `compound.spec.ts` was itself an instance of the bug.
- **Anchored regex patterns** (`@rapiq/core`): `createFilterRegexPattern` used `&&` instead of `&` for the CONTAINS check, so `startsWith`/`endsWith` produced **unanchored** patterns on regexp dialects — `startsWith(name, 'foo')` on pg emitted `"name" ~* 'foo'` and matched `xxfoo`, while the LIKE fallback on MSSQL correctly emitted `LIKE 'foo%'`. Patterns are now `^foo` / `foo$`; negated variants were already correct. (A pattern built without any anchor flag is now unanchored — previously it accidentally rendered `input$`.)
- **Empty `in`/`nin` lists** (`@rapiq/sql`): `Filter('in', field, [])` emitted the invalid SQL `field in()` — reachable from real client input through both parsers (`filter[id]=,` in parser-simple, `in(id)` in parser-expression). Now renders `1 = 0` (`nin`: `1 = 1`), with placeholder numbering kept continuous around it.
- **sqlite preset** (`@rapiq/sql`): was `{ ...mysql }`, inheriting a `regexp` callback although stock SQLite ships without a `REGEXP` function — `contains` emitted SQL that fails at runtime with `no such function: REGEXP`, while the LIKE fallback sat unreachable. The preset now omits `regexp`, routing anchored operators to LIKE (natively supported, incl. `ESCAPE`) and the `regex` operator to a typed `AdapterError`.

## Cleanups (same review)

- Fixed wrong operator type parameters on `visitFilterNotEndsWith` (`NOT_STARTS_WITH` → `NOT_ENDS_WITH`) and `visitFilterNotContains` (`CONTAINS` → `NOT_CONTAINS`).
- Collapsed the near-identical `visitFilterIn`/`visitFilterNotIn` bodies into a `whereIn` helper and the six anchored-operator bodies into a `whereAnchored` helper (matching the existing `whereLike` pattern), so future fixes land in one place.

## Tests

- New `packages/core/test/unit/parameter/filters-regex.spec.ts`: pattern + match assertions for all six flag combinations and case-insensitivity.
- `compound.spec.ts`: regression test for the precedence bug; complex-compound expectation updated to include the (correct) outer wrap.
- `in.spec.ts`: empty-list `in`/`nin` + placeholder continuity around them.
- `regex.spec.ts`: sqlite LIKE fallback + typed error, and anchored-pattern assertions for all six anchored operators on pg.

Docs (`integrations/sql.md`) updated for the sqlite preset and empty-list semantics; behavior log appended to `.agents/migration-notes.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected SQL handling for combined conditions, anchored string matching, and empty `in` / `nin` lists.
  * Improved fallback behavior for databases without regex support, including clearer errors for unsupported regex filters.
  * Fixed parenthesis placement in complex filter expressions.

* **Tests**
  * Added coverage for regex anchoring, negation, SQLite fallback behavior, and empty list conditions.

* **Documentation**
  * Updated SQL integration notes to reflect current filter behavior and dialect support.

### DIFF
--- a/.agents/migration-notes.md
+++ b/.agents/migration-notes.md
@@ -46,3 +46,11 @@ Known pre-existing issue (out of scope here, plan 006): `BaseParser.expandObject
 - **Error taxonomy** (plan 008 item 5): core exports `AdapterError` with `operatorUnsupported`/`featureUnsupported` factories and `ErrorCode.OPERATOR_UNSUPPORTED`/`FEATURE_UNSUPPORTED`; the raw `Error` throws in @rapiq/sql are migrated.
 - Relations adapter no longer duplicates parent path entries (`add('a.b'); add('a.c')` previously pushed `{path:'a'}` twice).
 - `AdapterOptions` accepts `rootAlias`, forwarded to the fields/filters/sort sub-adapters.
+
+## Post-review fixes (PR #741 audit)
+
+- **Compound wrapping**: `FiltersBaseAdapter.merge()` now parenthesizes by condition count instead of the `sql[0] !== '('` heuristic. Previously a nested compound whose first condition was itself parenthesized (e.g. the null-rewrite fragments) was merged unwrapped, producing wrong AND/OR precedence; a top-level multi-condition group is now always wrapped too.
+- **Anchored regex patterns** (`@rapiq/core`): `createFilterRegexPattern` used `&&` instead of `&` for the CONTAINS check, so `startsWith`/`endsWith` produced *unanchored* patterns on regexp dialects (pg/mysql/oracle) — they behaved like `contains`. Patterns are now anchored (`^foo`, `foo$`), matching the LIKE-fallback semantics on regexp-less dialects. A pattern built without an anchor flag is now unanchored (was: accidental `input$`).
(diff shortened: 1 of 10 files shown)


tada5hi:
@coderabbitai pause

PR #758, opened 2026-07-13
fix(sql): render negated filter operators null-inclusively

### LINKED ISSUE #752: Negated filter operators should match NULL rows in the SQL/TypeORM adapters (complement law)
## Problem

A negated filter condition evaluated through `@rapiq/sql` / `@rapiq/typeorm` silently excludes records whose field is `NULL`:

- `ne('name', 'Peter')` renders `name <> ?` — a row with `name IS NULL` is **not** returned.
- `nin('age', [20, 60])` renders `age NOT IN (...)` — a row with `age IS NULL` is **not** returned.
- `notContains` / `notStartsWith` / `notEndsWith` render a negated-lookahead regexp (or `NOT LIKE` on regexp-less dialects) — a `NULL` value yields `NULL`, so the row is **not** returned.

Expected behavior (settled 2026-07-08 with the `@rapiq/memory` design): **every negated operator is the exact logical complement of its positive twin** — a record that does *not* match `eq('name', 'Peter')` must match `ne('name', 'Peter')`, including records where the field is `NULL`/absent. The negations above should render null-inclusively, e.g. `(name <> ? OR name IS NULL)`.

Until then, the same `Query` disagrees across adapters: `@rapiq/memory` implements the complement law (pinned by its complement-law spec), while the SQL/TypeORM adapters follow raw three-valued logic. A guard using the memory predicate admits records the DB query would exclude, and vice versa. The divergence is currently documented in the memory integration docs as "alignment planned".

Parts of the rendering are **already complement-correct** and must not change:

- `ne(field, null)` → `field IS NOT NULL`
- `nin` with a null element → `(field NOT IN (...) AND field IS NOT NULL)` (De Morgan of `(IN ... OR IS NULL)`)
- `nin` with only null → `field IS NOT NULL`; `nin` with an empty list → `1 = 1`
- `exists(field, false)` → `field IS NULL`

## Root Cause Analysis

The filters visitor in the SQL adapter renders negated field operators as the bare SQL negation of the positive operator (`<>`, `NOT IN`, negated-lookahead regexp, `NOT LIKE`). Under SQL three-valued logic these comparisons evaluate to `NULL` (not `TRUE`) for `NULL` operands, so `NULL` rows fall out of the result — the SQL negation is *not* the set complement.

Contributing factors:

- The null-aware rewriting that already exists (the null-element `IN`/`NOT IN` handling, the `eq`/`ne` null special cases) only triggers when the *filter value* involves `null` — it never accounts for `NULL` in the *column*.
- The TypeORM adapter contains no rendering logic of its own; it reuses the SQL visitor wholesale, so it inherits the gap (and the fix) 1:1.
- The TypeORM test fixtures seed no record with a `NULL` value in any negated column, so the live-DB negation tests pass without ever exercising a `NULL` row.

## TDD Fix Plan

1. **RED**: SQL interpreter spec — `ne` with a non-null value renders a null-inclusive condition, e.g. `("name" <> $1 or "name" is null)` with the value as the only parameter; the `ne(null)` rendering stays `is not null`.
   **GREEN**: Rewrite the not-equal rendering to wrap the comparison with an `OR field IS NULL` branch.

2. **RED**: SQL interpreter spec — `nin` over a list *without* null elements renders `("field" not in(...) or "field" is null)`; the existing mixed-null (`AND IS NOT NULL`), only-null and empty-list renderings are unchanged.
   **GREEN**: Adjust the negated branch of the IN rendering so the no-null-element case appends the `OR IS NULL` clause.

3. **RED**: SQL interpreter spec — `notContains`/`notStartsWith`/`notEndsWith` render null-inclusively on a regexp dialect, e.g. `("name" ~* $1 or "name" is null)` with the negated-lookahead pattern parameter.
   **GREEN**: Wrap the negated anchored rendering (regexp path) with the `OR IS NULL` branch.

4. **RED**: Same for the LIKE fallback on regexp-less dialects (mssql/sqlite): `([name] not like ? escape '\' or [name] is null)`; positive anchored operators stay bare.
   **GREEN**: Wrap the negated LIKE rendering.

5. **RED**: TypeORM live-DB spec — seed a record with a `NULL` value in the filtered column; `ne`, `nin`, and a negated contains-family operator each return that record, while their positive twins do not.
   **GREEN**: Nothing beyond cycles 1–4 (TypeORM reuses the SQL visitor); the test pins the inheritance.

6. **RED**: Cross-adapter agreement spec — for a dataset containing `NULL`/absent fields, evaluating each negated operator via `@rapiq/memory` and via the TypeORM adapter on the live in-memory DB yields the same record set (the complement-law matrix the memory package already pins, replayed against the DB).
   **GREEN**: Nothing new if cycles 1–5 are complete; this is the acceptance gate for the alignment.

**REFACTOR**: Update the SQL integration docs (null-semantics and string-matching sections) to state that negations match `NULL` rows, and remove the "alignment planned" divergence row from the memory integration docs. Consider a shared helper for the `OR IS NULL` wrapping so all negated renderings stay consistent.

## Acceptance Criteria

- [ ] `ne`, `nin` (all list shapes) and `notContains`/`notStartsWith`/`notEndsWith` match `NULL` rows on every dialect preset (regexp and LIKE-fallback paths).
- [ ] Already-complement-correct renderings (`ne(null)`, null-element `nin`, empty-list `nin`, `exists(false)`) are unchanged.
- [ ] A TypeORM live-DB test seeds a `NULL` row and asserts the negated operators return it.
- [ ] The same `Query` produces the same record set via `@rapiq/memory` and via the TypeORM adapter for the complement-law input matrix.
- [ ] Docs updated: SQL page documents null-inclusive negations; memory page's divergence table no longer lists the gap.
- [ ] All new tests pass; existing tests updated where they pinned the old renderings.

tada5hi:
## Summary

Negated filter operators evaluated through `@rapiq/sql` / `@rapiq/typeorm` silently excluded records whose column is `NULL`: `ne('name', 'Peter')` rendered a bare `name <> ?`, which drops `NULL` rows under SQL three-valued logic. Every negated operator is now the exact logical complement of its positive twin — matching the semantics `@rapiq/memory` already pins with its complement-law spec.

A shared `whereComplement()` helper in the filters visitor wraps every negated rendering:

| Filter | SQL (before) | SQL (after) |
|---|---|---|
| `ne(field, a)` | `"field" <> $1` | `("field" <> $1 or "field" is null)` |
| `nin(field, [a, b])` | `"field" not in($1, $2)` | `("field" not in($1, $2) or "field" is null)` |
| `notContains` family (regexp dialects) | `"field" ~* $1` | `("field" ~* $1 or "field" is null)` |
| `notContains` family (LIKE fallback) | `[field] not like ? escape '\'` | `([field] not like ? escape '\' or [field] is null)` |

The already-complement-correct renderings are unchanged: `ne(null)` → `is not null`, `nin` with a null element → `(not in(...) and is not null)`, only-null / empty-list `nin`, and `exists(false)`.

The TypeORM adapter contains no rendering logic of its own, so it inherits the fix 1:1 through the shared visitor.

## Tests

- Updated the SQL interpreter specs that pinned the old renderings (`ne`, `nin`, negated anchored operators on both the regexp path and the LIKE fallback); the existing null-semantics spec stayed green throughout.
- New TypeORM live-DB specs seed a record with a `NULL` value in the filtered column and assert `ne`, `nin` and `notContains` return it while the positive twins do not.
- New cross-adapter agreement spec (`packages/typeorm/test/unit/complement.spec.ts`): the complement-law matrix (9 positive/negative pairs, including null-element / empty-list `in`/`nin` shapes) evaluated via `@rapiq/memory` and via the TypeORM adapter on a live better-sqlite3 database yields identical record sets, and each pair partitions the table. `@rapiq/memory` resolves through the workspace symlink — no dependency declaration needed for the test-only import.

## Docs

- `@rapiq/sql` page: null-semantics section documents the null-inclusive negations; string-matching section notes the LIKE-fallback form.
- `@rapiq/memory` page: removed the "alignment planned" divergence row.
- Filters guide: states the complement law under null semantics.

Closes #752

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL generation for negated filters (`ne`, `nin`, and negated string matching) so they are null-inclusive and correctly match records with missing or `NULL` field values.
  * Ensured consistent “complement” behavior across SQL adapters, aligning SQL results with in-memory filtering.

* **Documentation**
  * Updated null semantics documentation and added examples showing how negated operators behave with `NULL`/absent fields.

* **Tests**
  * Added and expanded unit coverage to verify complement behavior for null scenarios across adapters.

### DIFF
--- a/packages/docs/guide/filters.md
+++ b/packages/docs/guide/filters.md
@@ -97,6 +97,8 @@ The simple object/wire dialect only expresses flat **and** sets. For `or` and ne
 | `inArray(field, [a, null])` | `field IN (…) OR field IS NULL` |
 | `exists(field)` | `field IS NOT NULL` |
 
+Negated operators (`ne`, `nin`, `notContains`, `notStartsWith`, `notEndsWith`) are **exact complements** of their positive twins — they also match records where the field is `NULL`/absent. Adapters render them null-inclusively, e.g. `ne(field, a)` → `(field <> ? OR field IS NULL)`.
+
 ## Schema options
 
 ```typescript
--- a/packages/docs/packages/memory.md
+++ b/packages/docs/packages/memory.md
@@ -98,7 +98,6 @@ compileFilters(elemMatch('items', and(eq('id', 1), eq('active', true))))(user); 
 
 | Case | @rapiq/memory | Baseline |
 |---|---|---|
-| `ne`/`nin`/`not*` on absent values | match (complement law) | SQL renders `field <> ?`, which does **not** match `NULL` rows (alignment planned) |
 | Per-leaf array quantification | same-element binding | Mongo/ucast quantify each dotted condition independently |
 | `exists` | is-not-null | Mongo: property presence |
 | `contains` family | case-insensitive | Mongo/ucast: case-sensitive |
--- a/packages/docs/packages/sql.md
+++ b/packages/docs/packages/sql.md
@@ -94,6 +94,16 @@ A `null` filter value is rewritten to the SQL null predicates instead of being b
 
 An empty list never matches: `in(field, [])` renders `1 = 0` and `nin(field, [])` renders `1 = 1` (instead of the invalid `IN ()`).
 
+Negated operators are **exact complements** of their positive twins: a record that does not match `eq(field, a)` matches `ne(field, a)` — including records where the column is `NULL`. Since a bare SQL negation drops `NULL` rows under three-valued logic, negations render null-inclusively:
+
+| Filter | SQL |
+|---|---|
+| `ne(field, a)` | `(field <> ? OR field IS NULL)` |
+| `nin(field, [a, b])` | `(field NOT IN (...) OR field IS NULL)` |
+| `notContains(field, a)` (also `notStartsWith` / `notEndsWith`) | `(field ~* ? OR field IS NULL)` |
+
 ### String matching
 
 The `contains` / `startsWith` / `endsWith` operators (and their negations) match their value **literally** on every dialect: regex metacharacters are escaped on regex-capable dialects, LIKE wildcards are escaped on the LIKE fallback. Only the `regex` operator interprets its value as a pattern.
+
+The negations match rows whose column is `NULL` (complement law, see above) — on the LIKE fallback they render `(field NOT LIKE ? ESCAPE '\' OR field IS NULL)`.
--- a/packages/sql/src/visitor/filters.ts
+++ b/packages/sql/src/visitor/filters.ts
@@ -45,11 +45,16 @@ export class FiltersVisitor implements IFiltersVisitor<IFiltersAdapter>,
     }
 
     visitFilterNotEqual(expr: Filter<FilterFieldOperator.NOT_EQUAL>): IFiltersAdapter {
+        const field = this.adapter.buildField(expr.field);
         if (expr.value === null) {
-            return this.adapter.whereRaw(`${this.adapter.buildField(expr.field)} is not null`);
+            return this.adapter.whereRaw(`${field} is not null`);
         }
 
-        return this.adapter.where(expr.field, '<>', expr.value);
+        return this.whereComplement(
+            field,
+            `${field} <> ${this.adapter.buildParamPlaceholder()}`,
+            expr.value,
+        );
     }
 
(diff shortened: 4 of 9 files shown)


tada5hi:
@coderabbitai full review

PR #760, opened 2026-07-14
fix!: path-qualify join aliases for nested relations

### LINKED ISSUE #744: Join aliases collide for same-named relations on different branches
## Pre-existing issue found during PR review

**Reported in:** PR #743, review comments by @coderabbitai
**Files:** `packages/typeorm/src/adapter/relations.ts` (join dedup), `packages/sql/src/helpers/field.ts` (`parseField`)

## Description

Joins are aliased by the relation path's **last segment**: `role.realm` joins as alias `realm`. This is a system-wide convention — `parseField()` in `@rapiq/sql` resolves a field reference like `role.realm.name` to the prefix `realm`, so the filters/sort/fields adapters all reference joined columns via the leaf segment.

Consequently the typeorm `RelationsAdapter` dedupes joins by `joinAttribute.alias.name`. When two relation paths end in the same segment (e.g. `realm` and `role.realm` — both exist in the test entities), the later join is skipped and every `*.realm.*` field reference silently resolves against whichever join was applied first. Wrong results, no error.

Note: deduping by full relation path *without* changing the aliasing convention is not a fix — it would apply two joins with the same alias, producing ambiguous/invalid SQL. The alias itself must become path-qualified.

## Suggested fix

Coordinated change across `@rapiq/sql` and `@rapiq/typeorm`:

- Alias joins by the full relation path (e.g. `role_realm` for `role.realm`).
- Change `parseField()` (and the field-prefix handling in the filters/sort/fields base adapters) to emit the matching path-qualified prefix instead of the leaf segment.
- Keep dedup keyed by the path-qualified alias; pre-existing consumer joins matched by alias keep working.
- Update the alias-convention caveat in `packages/docs/integrations/typeorm.md` once landed.

This changes generated SQL aliases for nested relations, so it should land as its own breaking change.

tada5hi:
## Summary

Fixes #744.

Joins were aliased by the relation path's **last segment**: `role.realm` joined as alias `realm`, the same convention `parseField()` resolved field references against. Two relation paths ending in the same segment (e.g. `realm` and `role.realm`) therefore collided — the typeorm relations adapter deduped by alias, skipped the later join, and every `*.realm.*` field reference silently resolved against whichever join was applied first. Wrong results, no error.

## Changes

- **`@rapiq/sql`**: join aliases are now derived from the **full relation path** with `.` → `_` (`role.realm` → `role_realm`). The derivation lives in one exported helper, `buildRelationAlias(path)`, owned at runtime by the shared relations adapter — `parseField()` and the fields/filters/sort base adapters resolve field prefixes through it, so join application and column references can never disagree again.
- **`@rapiq/typeorm`**: the join walk chains path-qualified aliases (`applyJoin('role.realm', 'role_realm')`) and dedupes by the derived alias. Single-segment paths keep their leaf alias, so pre-existing consumer joins on direct relations (`leftJoin('user.realm', 'realm')`) are still recognized.
- **Escape hatch**: the derivation is injectable via `relationAlias` (sql `AdapterOptions` / typeorm `relations` options) — for custom separator conventions, identifier length limits, or matching pre-existing nested joins.
- **Docs**: alias-convention warnings rewritten on the `@rapiq/sql` and `@rapiq/typeorm` pages.

## Tests

- sql: fragments regression for `realm` vs `role.realm` (fields + filters + sort), custom `relationAlias` derivation; updated nested elemMatch/auto-join expectations to the path-qualified prefixes.
- typeorm: join-attribute regression asserting distinct aliases for same-named relations on different branches, custom derivation, and an end-to-end sqlite hydration test where `user.realm` and `user.role.realm` hold **different** realms — fails on the old aliasing by construction.

## Breaking change

Generated SQL aliases for nested relation paths are now path-qualified (`role.realm` joins as `role_realm`, not `realm`). Consumer-applied joins for nested paths are only recognized (deduped) under the path-qualified alias; direct (single-segment) relation joins are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added consistent, path-qualified aliases for nested relationship joins, preventing collisions between similarly named relationships.
  * Added an option to customize relationship alias generation, including support for database identifier constraints.
  * Updated field, filter, sort, and join handling to use the same alias convention.

* **Documentation**
  * Documented default and custom relationship alias behavior for SQL and TypeORM adapters.

* **Bug Fixes**
  * Improved nested relationship queries so fields, filters, sorting, and joins resolve to the correct relationships.


tada5hi:
@coderabbitai full review

PR #762, opened 2026-07-14
feat!: case-insensitive string equality with per-field opt-out

tada5hi:
## What

String equality (`eq` / `ne` / `in` / `nin`) now matches **case-insensitively by default, uniformly across every backend** — aligning the equality family with the anchored operators (`contains`/`startsWith`/`endsWith`), which have been case-insensitive since v2. `eq('name', 'super hero')` matches `Super Hero` on Postgres, MySQL, in `@rapiq/memory`, and through `@rapiq/typeorm` alike.

Previously, equality delegated case behavior to the database: the same query matched different records on MySQL (`*_ci` collation) than on Postgres — exactly the cross-backend non-determinism rapiq exists to eliminate.

## How

- **`@rapiq/sql`**: string equality folds both comparison sides through a new `caseFold` dialect callback — `lower(field) = lower(?)`, `lower(field) IN (lower(?), ?)` (only string members fold; null semantics and the complement law are unchanged).
- **Dialect presets**: `mysql` and `mssql` set `caseFold` to identity — their default `*_ci` collations already compare case-insensitively, and skipping `lower()` keeps plain indexes usable. `pg`/`sqlite`/`oracle` use the `lower()` default.
- **`@rapiq/memory`**: string conditions compare via `toLowerCase()`, mirroring the SQL rendering; complement law preserved (`ne`/`nin` wrap the same tests).
- **`@rapiq/typeorm`**: inherits the shared SQL visitor; resolves `caseFold` from the detected dialect. Covered by live better-sqlite3 round trips.
- **Column-type detection (typeorm)**: a new `isCaseFoldable(field)` adapter hook (default: fold) gates folding by column type — the typeorm adapter resolves fields against entity metadata (relation paths included) and folds only string-typed columns, so an int column filtered with an untyped wire string (`filter[age]=18`) renders a plain `=` instead of a `lower(integer)` type error on postgres.
- **Opt-out**: new `filters.caseSensitive` schema list for fields where exactness matters (ids, tokens); forwarded to adapters via `execute(query, { visitor: { caseSensitive } })` (sql/typeorm) or `{ filters: { caseSensitive } }` (memory). Non-string values (numbers, booleans, dates, `null`) are untouched; range comparisons and sort ordering stay collation-governed.

## Docs

Filters guide gained a *Case sensitivity* section (incl. Postgres expression-index guidance), schema options extended, `@rapiq/sql` + `@rapiq/memory` package pages updated, and a breaking-change entry added to the v1 migration guide.

## BREAKING CHANGE

`eq`/`ne`/`in`/`nin` match strings case-insensitively on every adapter; previously case behavior followed the database collation (case-sensitive on postgres/sqlite/oracle and in `@rapiq/memory`). Opt identifier/token fields out via `filters.caseSensitive`.

## Summary by CodeRabbit

* **New Features**
  * String equality and membership filters are now case-insensitive by default across supported adapters.
  * Added per-field `caseSensitive` configuration for exact string matching.
  * Preserved typed comparisons for non-string values and backend collation behavior for ranges and sorting.

* **Documentation**
  * Added guidance on case-sensitivity behavior, configuration, migrations, and adapter-specific considerations.

* **Tests**
  * Added coverage for default matching, opt-outs, mixed values, null handling, and SQL dialect behavior.


tada5hi:
@coderabbitai pause

PR #770, opened 2026-07-17
feat!: add ITSELF self-reference marker for element-level $elemMatch and $all

### LINKED ISSUE #768: Feature: ITSELF self-reference marker (element-level $elemMatch + $all)
## Motivation

`@rapiq/parser-mongo` currently rejects two mongo constructs with typed errors because the core IR cannot express them:

- element-level `$elemMatch` (`{ scores: { $elemMatch: { $gt: 5 } } }`) → `FEATURE_UNSUPPORTED`
- `$all` (`{ tags: { $all: ['a', 'b'] } }`) → `OPERATOR_UNSUPPORTED`

Both were deliberate scope cuts in the parser-mongo design (plan 013, Q5/Q8): the IR has no way to reference *the array element itself* as the field of a leaf condition. ucast solves this with an `ITSELF` sentinel (see `.agents/references/ucast.md`).

Consumer demand has now appeared: migrating `authup/access` off `@ucast/mongo2js` onto `@rapiq/parser-mongo` + `@rapiq/memory` — user-authored attribute policies may contain either construct, and today they fail at parse instead of evaluating.

## Why `$all` cannot be desugared without this

`$all: [a, b]` means "for **each** listed value, **some** array element equals it" — independent existential quantification per value.

- `and(eq(f, a), eq(f, b))` is wrong: under the settled same-element join-row binding (plan 014 Q4; `packages/memory/src/parameter/filters/binding.ts`, SQL single-join-alias parity), all conditions on the same dotted path share one element binding — one element would have to equal both values, so the AND is unsatisfiable for distinct values.
- `or(eq(f, a), eq(f, b))` is `$in` semantics ("contains a **or** b"), not `$all`.
- On flat scalar-array leaves, `eq` compares the array value itself, not its elements.

With a self-reference marker, `$all` needs **no new `FilterFieldOperator`** — the parser desugars it to an AND of independently-scoped element matches:

```
$all: [a, b]  ≡  and(
    elemMatch(f, eq(ITSELF, a)),
    elemMatch(f, eq(ITSELF, b)),
)
```

and the element-level `$elemMatch` form desugars the same way (`{ $elemMatch: { $gt: 5 } }` ≡ `elemMatch(f, gt(ITSELF, 5))`).

## Design sketch

- **core**: a reserved self-reference marker legal in the `field` position of a leaf `Filter` **inside an `elemMatch` interior only** (elsewhere → parse/build error). Key design question: representation. ucast uses a symbol, but a rapiq marker must survive codec round-trips (expression dialect wire form) and key-path resolution — a reserved token string is the likely shape; exact spelling to be settled here.
- **parser-mongo**: accept the element-level `$elemMatch` operator form (lifting the Q8 throw) and map `$all` via the desugar above (`$all` value validation mirroring `$in`: non-empty array of scalar|null|Date). Negated `$all` (under `$not`/`$nor`): no complement twin — throw `OPERATOR_UNSUPPORTED`, consistent with `$regex`/`$mod`/`$elemMatch` (plan 013 Q3).
- **memory**: evaluate an ITSELF leaf against the bound element itself — the binding machinery already enumerates elements; the leaf lookup just short-circuits property resolution.
- **sql / typeorm**: a joined relation row is not a scalar column — throw typed `AdapterError.featureUnsupported` initially (precedent: mssql regex). Dialect-level JSON-array support (`json_each`/`unnest`) can be a follow-up.
- **codec-url (expression dialect)**: the grammar needs a token for the self-reference in `elemMatch(...)` interiors; simple dialect cannot express it → encode throws under the subset law, as designed.
- **build layer**: decide whether the condition helpers expose the marker (e.g. exported constant usable as `eq(ITSELF, v)` inside `elemMatch`).

## Acceptance criteria

- [ ] Core exports the self-reference marker; misuse outside an `elemMatch` interior fails typed.
- [ ] `MongoFiltersParser` parses `{ f: { $elemMatch: { $gt: 5 } } }` and `{ f: { $all: [...] } }` into the AST shapes above.
- [ ] `@rapiq/memory` evaluates both with mongo semantics (element-wise, independent quantification per `$all` value; missing/non-array field → no match).
- [ ] `@rapiq/sql` / `@rapiq/typeorm` throw `AdapterError.featureUnsupported` for ITSELF leaves.
- [ ] Expression codec encodes/decodes ITSELF round-trip; simple codec encode throws typed.
- [ ] Docs: mongo dialect page drops `$all`/element-level `$elemMatch` from the unsupported list; memory + sql pages document the semantics/limitation.

tada5hi:
Closes #768.

The reserved marker `ITSELF` (wire spelling `$this`) addresses the array element itself in the field position of a leaf condition inside an `elemMatch` interior — the core IR gap behind the two remaining mongo constructs `@rapiq/parser-mongo` had to reject. Design decisions (marker spelling, binding scope) are recorded in `.agents/plans/016-itself-marker.md`.

## Changes

- **core**: exports `ITSELF = '$this'`; `defineFilters` validates the marker contract on condition-node input, accepts mongo-style element-level `$elemMatch` operator objects (`{ $elemMatch: { $gt: 5 } }`), and now rejects `$`-prefixed keys as field names (matching the mongo-dialect reservation).
- **parser-mongo**: lifts the plan-013 Q8 throw — element-level `$elemMatch` parses onto `ITSELF` (ucast `hasOperators` split, including `$not` and nested `$elemMatch` interiors); `$all` moves out of the unsupported set and desugars to an AND of independently scoped `elemMatch(f, eq(ITSELF, v))` conditions with `$in`-style value validation; negated `$all`/`$elemMatch` throw `OPERATOR_UNSUPPORTED`.
- **memory**: every `elemMatch` node opens its own quantifier scope (discriminated binding paths) — interiors share one element, sibling elemMatches bind independently (exact Mongo semantics, required for the `$all` desugar). `ITSELF` leaves evaluate the bound element and only match real array elements: missing fields, scalars, to-one objects and empty arrays never match (deliberately no mongo scalar fallback for `$all`).
- **sql / typeorm**: `FiltersBaseAdapter.buildField` throws typed `AdapterError.featureUnsupported('filters:itself')` — a joined relation row is not a scalar column; dialect JSON-array support (`json_each`/`unnest`) stays a follow-up.
- **parser-expression / codec-url**: new `elemMatch(field, expr)` grammar production (interior fields resolve against the related schema via `descend`, unbound fallback for non-relation fields) and a dedicated `$this` token; `not(elemMatch(...))` throws. The expression encoder serializes `ELEM_MATCH` with full round-trip coverage — wire form `filter=elemMatch(scores,gt($this,'5'))`; the simple codec keeps throwing typed under the subset law.
- **docs**: parser-mongo (unsupported list updated, `$all` row), memory (quantifier semantics + ITSELF section, divergence table), sql (ITSELF limitation), parser-expression (grammar), codec-url and the filters / building-queries / wire guides.

## Breaking changes

- Two `elemMatch` nodes on one field (or an `elemMatch` beside a dotted sibling) no longer share an element binding in `@rapiq/memory` — the two plan-014 parity-fallout pinning specs were flipped. SQL keeps single-join-alias composition, so multi-elemMatch trees diverge between sql and memory (documented).
- `$`-prefixed keys are rejected as field names in the build-layer filters object grammar.
- In the expression filter dialect, `elemMatch` is a reserved keyword and `$`-words are reserved markers — fields with these names no longer tokenize/encode.

## Verification

All 8 packages build; full test run passes (~60 new specs across core, parser-mongo, parser-expression, memory, sql, typeorm, codec-url); lint clean. End-to-end smoke: a mongo document with `$all` + element-level `$elemMatch` parses, evaluates with mongo semantics via `compileFilters`, and round-trips the URL codec.

## Summary by CodeRabbit

* **New Features**
  * Added `elemMatch` support to expression and Mongo-style query formats.
  * Added `$all` support for independently matching array elements.
  * Added `$this` element references for conditions inside `elemMatch`.
  * Added in-memory evaluation for nested and element-level array matching.

* **Bug Fixes**
  * Corrected matching so separate `elemMatch` conditions evaluate independently.

* **Documentation**
  * Expanded guidance on array matching, `$this`, dialect support, and adapter limitations.

### DIFF
--- a/.agents/references/ucast.md
+++ b/.agents/references/ucast.md
@@ -15,7 +15,7 @@ package is the reference implementation for rapiq's mongo parser dialect
 | `DocumentCondition` (e.g. `$where`) | — | rapiq has no document-level conditions (and `$where` is a code-execution foot-gun; deliberately unsupported) |
 | `NULL_CONDITION` sentinel (dropped by `pushIfNonNullCondition`) | — | rapiq parsers just don't emit a node (e.g. `$options` is consumed by its `$regex` sibling) |
 | `optimizedCompoundCondition()` (`core/src/utils.ts`) — flattens same-operator nesting, unwraps single-child compounds | `Filters.flatten()` | rapiq's flatten does not unwrap single-child compounds |
-| `ITSELF` field sentinel (elemMatch on scalar arrays) | — | rapiq `elemMatch` values are conditions with fields relative to the array element |
+| `ITSELF` field sentinel (elemMatch on scalar arrays) | `ITSELF = '$this'` (`core/src/parameter/filters/constants.ts`, plan 016) | ucast uses a symbol; rapiq uses a reserved token string so the marker survives codec round-trips (expression wire form `elemMatch(scores,gt($this,'5'))`) |
 
 ## Parsing (`@ucast/core` ObjectQueryParser + `@ucast/mongo`)
 
@@ -38,10 +38,14 @@ package is the reference implementation for rapiq's mongo parser dialect
 - `$regex` + sibling `$options`: string is compiled to `new RegExp(value, $options)`; `$options`
   itself parses to `NULL_CONDITION` (consumed by the `$regex` instruction via `context.query`).
 - `$elemMatch`: `hasOperators(value)` ? parse as field operators on `ITSELF` : parse as nested
-  query; result nested inside a `FieldCondition`.
+  query; result nested inside a `FieldCondition`. rapiq mirrors the split since plan 016
+  (`MongoFiltersParser.buildElemMatch`: any `MONGO_FIELD_OPERATORS` key → element-level form on
+  `ITSELF`).
 - Value validation throws plain `Error` (`ensureIs*` helpers); rapiq uses typed
   `FiltersParseError` + `ErrorCode` members instead.
-- Operators with no rapiq AST equivalent: `$size`, `$all`, `$where`, `$type`.
+- `$all`: ucast keeps it as its own condition; rapiq desugars it to an AND of independently
+  scoped `elemMatch(f, eq(ITSELF, v))` conditions (plan 016 Q4, no dedicated operator).
+- Operators with no rapiq AST equivalent: `$size`, `$where`, `$type`.
 
 ## Interpretation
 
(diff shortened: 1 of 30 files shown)


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #780, opened 2026-07-19
feat(typeorm): derive schemas from entity metadata

tada5hi:
## Summary

New `schema/` module in `@rapiq/typeorm`: derive rapiq schemas and their registry linkage from TypeORM `EntityMetadata` instead of hand-maintaining them.

- **`defineSchemaWithEntity(target, dataSource, options?)`** — also accepts a built `EntityMetadata` directly. Structure is always derived: lower-camel schema name (`RoleDetail` → `roleDetail`), `relations.allowed` from relation property names, `schemaMapping` from each relation's `inverseEntityMetadata`.
- **Column allow-lists are opt-in** per parameter via the `allowed: 'columns'` sentinel — expands to the entity's column property paths. Hidden columns (`select: false`) and virtual join columns are always excluded; explicitly declared FK columns (e.g. `realm_id`) are included. Explicit options always win over derived values.
- **`defineSchemaRegistryWithDataSource(dataSource, { schemas?, registry? })`** — one schema per entity (junction/closure-junction metadata skipped), cross-linked by construction so nested paths like `role.detail` resolve without manual wiring. Per-entity `schemas` options are keyed by derived name or by entity class (`Map`); an options key matching no entity throws, so renames fail loudly.
- **Extending an existing registry**: pass `registry` to fill in derived schemas around hand-written ones — already-registered names take precedence (the entity is skipped), and `schemas` options targeting a skipped name throw instead of being silently ignored.
- Derivation never sets `strict`; the docs call out the `strict: true` + `allowed: 'columns'` interplay explicitly.

```typescript
const registry = defineSchemaRegistryWithDataSource(dataSource, {
    schemas: {
        user: {
            filters: { allowed: 'columns' },
            sort: { allowed: 'columns' },
        },
    },
});
```

Docs: new "Deriving schemas from entities" section in `packages/docs/packages/typeorm.md`.

Context: this is the successor feature to typeorm-extension's rapiq integration, which is being removed in its v4 (tada5hi/typeorm-extension#1411).

## Notes

- Registry failure modes (duplicate derived name, unknown/inapplicable options key) currently throw plain `Error`; a typed error class is a follow-up.
- Entity decorators were considered and deliberately deferred — metadata + the options object cover the same ground without the `experimentalDecorators` toolchain friction.

## Tests

13 new specs in `packages/typeorm/test/unit/schema/module.spec.ts` — derivation rules, precedence, both registry keying styles, registry extension, and end-to-end cross-schema relation traversal through `SimpleParser`. Edge cases (`select: false`, virtual join columns) use a spec-local entity; the shared fixtures are untouched. Package suite: 88/88.

## Summary by CodeRabbit

* **New Features**
  * Added enhanced TypeORM integration to automatically derive schemas and build schema registries from entity metadata.
  * Supports derivation of schema names, relation mapping, and column-based allow-lists (with an opt-in “inherit” mechanism).
  * Added entity-targeted schema options, override precedence, and the ability to extend existing registries safely.
* **Documentation**
  * Added a new guide section covering deriving schemas from entities, allow-list rules, override behavior, and required metadata setup.
* **Tests**
  * Added unit tests covering derivation, registry behavior, overrides, and error cases.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4137,7 +4137,6 @@
             "version": "5.4.4",
             "resolved": "https://registry.npmjs.org/change-case/-/change-case-5.4.4.tgz",
             "integrity": "sha512-HRQyTk2/YPEkt9TnUPbOpr64Uw3KOicFWPVBb+xiHvd6eBx/qPr9xqfBFDT8P2vWsvvz4jbEkfDe71W3VyNu2w==",
-            "dev": true,
             "license": "MIT"
         },
         "node_modules/character-entities-html4": {
@@ -9926,6 +9925,9 @@
             "name": "@rapiq/typeorm",
             "version": "2.0.0-beta.0",
             "license": "MIT",
+            "dependencies": {
+                "change-case": "^5.4.4"
+            },
             "devDependencies": {
                 "@rapiq/core": "^2.0.0-beta.0",
                 "@rapiq/parser-simple": "^2.0.0-beta.0",
--- a/packages/docs/packages/typeorm.md
+++ b/packages/docs/packages/typeorm.md
@@ -88,8 +88,8 @@ import { defineSchemaRegistryWithDataSource } from '@rapiq/typeorm';
 const registry = defineSchemaRegistryWithDataSource(dataSource, {
     schemas: {
         user: {
-            filters: { allowed: 'columns' },
-            sort: { allowed: 'columns' },
+            filters: { allowed: 'inherit' },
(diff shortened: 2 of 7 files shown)


tada5hi:
Downstream consumer waiting on this: authup/authup#3279 (Phase 2 — evaluate metadata-derived base schemas with explicit allow-list overlays).

tada5hi:
@coderabbitai review

PR #807, opened 2026-07-20
feat(core): request-context threading & per-parameter schema validate hooks

### LINKED ISSUE #806: RFC: request-context threading & per-parameter validate hooks (dynamic authorization at decode time)
## Problem

A `Schema` expresses what a client may request as **static, per-process allow-lists**. Anything that depends on *who is asking* has exactly one seam today, and it is not enough:

- **`filters.validate` is the only dynamic hook** — `Validator = (input: IFilter) => MaybeAsync<IFilter | undefined>` (`packages/core/src/schema/parameter/filters/types.ts:24`). It is filters-only and receives **no context**: a per-request actor can only reach it by closure, which breaks shared per-process schema instances.
- **Relations, fields and sort verdicts are purely static.** Every parser routes them through `ResolutionScope.checkName` / `checkSegment` (`packages/core/src/schema/resolver/module.ts`), which consult the allow-list/mapping and nothing else.
- **`fields.verify` is a phantom.** `FieldsOptions.verify?: VerifyFn<string[], CONTEXT>` (`packages/core/src/schema/parameter/fields/types.ts`) even declares a context-carrying hook shape — but nothing ever invokes it.
- **No option type carries a caller context.** `ParseQueryOptions`, `ParseParameterOptions` (`packages/core/src/parser/types.ts`), and the backend `ExecuteOptions` all lack one; the only slot is the untyped `[key: string]: any` index signature on @rapiq/sql's `VisitorOptions`.

### Motivating failure

A junction-entity endpoint (real case: authup's ClientPermission service) gates the list operation on the junction's own permissions and lets the schema allow `include=client` / `include=permission`. The permission actually required to *read the included entities* (`client_read` / `permission_read`) is checked nowhere — not at decode time, not on the response — because the application has no way to run "may **this actor** include **this relation**?" inside the parse. An actor holding only junction permissions reads full joined rows via `?include=`.

This is the include-authorization seam every comparable layer ends up needing: feathers-casl runs an `authorize` before-hook that narrows the query per request; Hasura authorizes every nesting level against the *target* table's own per-role permission; spatie/laravel-query-builder (static `allowedIncludes`, like us) has role-dependent narrowing as a recurring feature request.

### 1. A `context` parse/decode option

```ts
parser.parse(input, { schema: 'user', context: actor });
await codec.decodeAsync(req.query, { schema: 'user', context: actor });
```

- `ParseQueryOptions` gains `context?: CONTEXT`. `URLCodec.decode`/`decodeAsync` forward `ParseQueryOptions` verbatim, so the codec entry points get it for free.
- `prepareQueryContext` copies it into the per-parameter options, and it lands in `ResolutionScopeContext` — which is **inherited into child scopes on `descend`**, so one seam covers arbitrarily deep `include=client.realm` paths.
- The context is opaque to rapiq: `CONTEXT` is a defaulted generic — `Schema<RECORD, CONTEXT = unknown>`, `SchemaRegistry<CONTEXT = unknown>`, `defineSchema<RECORD, CONTEXT>` — non-breaking, one context type per registry (matches the one-actor-type-per-app reality). Registries with mixed schemas erase to `unknown`.

### 2. Per-parameter validate hooks receiving the context

```ts
const userSchema = defineSchema<User, Actor>({
    relations: {
        allowed: ['realm', 'items'],
        validate: (relation, actor) => actor.can(`${relation}_read`),   // MaybeAsync<boolean | undefined>
    },
    filters: {
        validate: (filter, actor) => filter,                             // arity extension, backward compatible
    },
});
```

- **`relations.validate?: (name: string, context: CONTEXT) => MaybeAsync<boolean | undefined>`** — invoked at the `ResolutionScope` segment verdict, i.e. once per resolved segment during descent: `include=client.realm` fires `client` against the root schema and `realm` against the client schema (via `schemaMapping`), matching the "authorize against the target schema, not the include path" model. `false`/`undefined` vetoes.
- **`fields.validate` / `sort.validate`** — same shape, wired at the `checkName` verdict. The phantom `fields.verify` is removed in the same change.
- **`filters.validate` gains the context as a second argument** — `(input: IFilter, context: CONTEXT) => MaybeAsync<IFilter | undefined>`. Existing one-arg validators keep working.

### 3. Veto semantics: the existing dial

A hook veto follows the same drop-vs-throw policy as an allow-list miss: **drop by default, `throwOnFailure` opts into throwing** — no second failure model. Thrown form gets a dedicated `KeyResolutionErrorCode` member (working name `VALIDATE_REJECTED`) surfaced through the existing per-parameter error classes, so callers can distinguish "not allowed by schema" from "rejected by validator".

### 4. Async: the established split

Same contract as `filters.validate` today: the sync `parse()` throws `SCHEMA_VALIDATOR_ASYNC_REQUIRES_ASYNC_PARSER` when a hook returns a thenable (`packages/core/src/parser/parameter/filters/validate.ts`); `parseAsync()`/`decodeAsync()` await. Permission checks are async in practice, so consumers use the async entry points.

## Non-goals (follow-ups, not this RFC)

- **Relation-scoped residual conditions** — attaching a per-relation condition so unauthorized *rows* of an included relation come back `null` (join-ON rendering in sql/typeorm, sub-object predicates in memory; Hasura's nested-object model). This RFC's hook is deliberately coarse (include allowed or not); row-level narrowing is a separate IR extension and will get its own RFC.
- **`ExecuteOptions.context` on backends** — deferred until a concrete consumer exists.
- **Any policy vocabulary in rapiq.** rapiq stays IR + hooks + interpreters; policy engines compile *their* trees into conditions/predicates on top (`query.filters.and(...)` injection and `@rapiq/memory` `compileFilters` already cover the in-query and post-query phases).

## Open questions

1. **Async mechanics.** Asyncify `ResolutionScope` resolution (invasive), or keep resolution sync and collect hook invocations as *obligations* resolved after key resolution (per-parameter parsers await verdicts, prune vetoed nodes)? Leaning obligations — smaller blast radius, keeps scope logic sync.
2. Does `sort.validate` earn its keep in phase 1, or ship relations+fields+filters only?
3. Error-code naming (`VALIDATE_REJECTED` vs something clearer).

## Docs impact

`guide/recipes/authorization.md` (context + hooks become the first layer of the layered-defense table), schema API pages, and the parser/codec pages for the new option.

---

Companion RFC on the consuming side: authup/authup#3286 — data-availability-driven policy evaluation (tri-state engine), whose derived pre-gate is what a `relations.validate` hook calls per include.

tada5hi:
Implements RFC #806, phase 1: per-request dynamic authorization at decode time.

## What

- **`context` parse/decode option** — `ParseQueryOptions`/`ParseParameterOptions` carry an opaque `context?: unknown` (typically the authenticated actor). `URLCodec.decode/decodeAsync` forward options verbatim, so the codec entry points get it for free.
- **Per-key `validate` hooks** on the `relations`, `fields` and `sort` sub-schemas — `(name, context) => MaybeAsync<boolean | undefined>`. Hooks run on the canonical (alias-resolved) key against the schema that governs it: `include=items.realm` invokes the root schema's hook with `items` and the item schema's hook with `realm` (via `schemaMapping`), so an include can never bypass the related schema's own gate. A rejected relation also prunes every deeper relation reached through it.
- **`filters.validate` arity extension** — the existing leaf validator now receives the context as a second argument (backward compatible; all three dialects forward it).
- **Failure policy** — hook rejections follow the existing dial: dropped by default, thrown under `throwOnFailure` with the new `ErrorCode.KEY_VALIDATE_REJECTED` (distinguishable from allow-list misses).
- **Sync/async** — mirrors the filters-validator contract: hooks are collected during resolution and evaluated afterwards; sync `parse()` refuses a thenable with `SCHEMA_VALIDATOR_ASYNC_REQUIRES_ASYNC_PARSER`, `parseAsync()`/`decodeAsync()` await sequentially.
- **Typing** — `Schema`, `SchemaRegistry`, `defineSchema` and the per-parameter options/factories carry a defaulted `CONTEXT` generic, so hooks are typed at the definition site (`defineSchema<User, Actor>`).
- **Removed** the never-invoked `fields.verify` option and its `VerifyFn` type (declared with a context parameter but dead since introduction).

## Design decisions

- **Obligations over an async `ResolutionScope`.** Key resolution stays fully synchronous; parsers record `(key, path, schema)` obligations during resolution and evaluate them once the parameter is assembled (`applyKeySchemaValidation`/`...Async` in core). One resolution pass serves both entry points, and pruning is name-based since `Fields.execute` clones nodes.
- **Client input only.** Schema `default`s are server-authored and bypass the hooks; rejected keys are removed after parameter assembly, so `fields`/`sort` defaults do not re-materialize when a hook empties the selection (documented).
- **`CONTEXT` defaults to `any`, not `unknown`** — under strict function-type contravariance, `unknown` would make `Schema<User, Actor>` unassignable to the bare `Schema<RECORD>` references used internally (resolver, registry, parsers).
- **Sort hooks run after tuple-group matching** — a vetoed member drops from a matched group (documented).

## Out of scope (per the RFC)

Relation-scoped residual conditions (join-ON rendering), `ExecuteOptions.context` on backends, and any policy vocabulary — rapiq stays IR + hooks + interpreters.

## Docs

- `guide/schemas.md`: new "Validate hooks & parse context" section + per-parameter option table rows
- `guide/recipes/authorization.md`: new "Gating: per-actor checks at decode time" layer (motivating include-permission case) + layered-defense row
- `guide/filters.md`, `guide/errors.md`: context argument, `KEY_VALIDATE_REJECTED`

## Tests

- New `parser-simple` spec: relation drop/keep with context, `undefined` context default, target-schema resolution for deep paths, descendant pruning, `throwOnFailure`, sync-refuses-async, `parseAsync` awaiting, fields/sort veto + defaults-bypass, filters context forwarding, end-to-end context threading through `SimpleParser`.
- Full monorepo: 9/9 build, 8/8 test targets green (uncached), lint clean on changed files, docs site builds.

Closes #806

## Summary by CodeRabbit

* **New Features**
  * Added per-key validation hooks for fields, relations, and sort parameters.
  * Parse and decode operations can now pass custom context to validation hooks.
  * Rejected keys are automatically removed, with an option to raise a parse error instead.
  * Added support for asynchronous validation during async parsing.
  * Filter validation hooks now receive the supplied context.

* **Documentation**
  * Added guidance and examples for context-aware validation and authorization scenarios.
  * Documented the new key-validation error code and rejection behavior.


tada5hi:
@coderabbitai pause

PR #816, opened 2026-07-21
feat(core): authorize relation paths traversed by filters/fields/sort (#815)

### LINKED ISSUE #815: relations.validate bypassed for relation paths referenced only by filters/fields/sort (SQL adapters auto-join)
## Context

Follow-up to #806 / #807 (shipped in core 2.0.0-beta.5; verified on beta.6). authup is wiring decode-time include authorization via `RelationsOptions.validate` (authup/authup#3295): the hook gates each `include=` segment on the acting identity's read permission for the relation's target entity type.

## Problem

The relations validate hook only runs for keys of the **relations parameter** (`include=`). A relation path referenced *only* by another parameter never invokes it:

```ts
// relations: { allowed: ['user'], validate: denyHook }, schemaMapping: { user: 'user' }
decode({ filter: { 'user.name': 'admin' } }, { schema: 'userRole', context })   // hook NOT called, filter survives
decode({ fields: { user: 'email' } },        { schema: 'userRole', context })   // hook NOT called, fields keeps 'user.email'
decode({ sort: 'user.name' },                { schema: 'userRole', context })   // hook NOT called, sort survives
```

Meanwhile the SQL adapters **auto-join** such paths: `FieldsAdapter` / `FiltersAdapter` / `SortAdapter` all do `if (output.relation) this.relations.add(output.relation)`, and `RelationsAdapter.execute()` joins everything added. So the join (and with a dotted `fields[...]` selection even target-entity **data**) is reachable without the relations gate ever running.

Contrast: when the include IS requested and the hook rejects it, the pruning correctly cascades — dotted filter/sort/field keys on the denied path are dropped too (verified on beta.6). The gap is only the include-less path.

## Impact (authup)

An actor authorized for a junction list (e.g. `user_role_read`) but lacking the member's read permission (`user_read`) is stripped of `include=user` by the gate — but can still:

- probe existence via `filter[user.name]=...` (join + WHERE, no gate), and
- with `fields[user]=email`, materialize target-entity columns through the auto-joined alias.

## Proposal

Run the relations validate hook for every relation path a parameter resolves through — not just paths of the relations parameter. I.e. when fields/filters/sort resolution walks a dotted key through `schemaMapping`, record the traversed relation segments as `PendingKeyValidation` obligations against the governing schemas' **relations** schema (dedup'd with the include-driven obligations), and prune the dependent keys when rejected. That keeps a single authorization point (the relations hook) covering everything that can force a join downstream.

Until then authup documents the residual with this issue as the removal trigger.

tada5hi:
## Summary

Closes #815.

`relations.validate` (the per-request relation authorization hook) only ran for the **relations** parameter. A relation referenced *only* by a dotted `filter` / `fields` / `sort` key traverses `schemaMapping` and is **auto-joined** by the `@rapiq/sql` and `@rapiq/typeorm` adapters, so the gate never ran. An actor denied `include=user` could still probe or materialise `user` via `filter[user.name]`, `fields[user]=email` or `sort=user.name`.

## Approach

Authorization is enforced at parse/decode time by pruning the AST, so **every** backend (sql, typeorm, memory) consumes an already-safe `Query` — not per-adapter.

- **Choke-point recording.** Relation-authorization obligations are recorded inside `ResolutionScope.resolveKey`, the single point every dialect resolves a key through. A relation is authorized iff a leaf under it resolves — i.e. iff it is actually used. Robust by construction: a new operator or dialect can't reach a relation without going through the choke point.
- **Single pooled pass.** Each parameter records into a per-request ledger; `BaseQueryParser` evaluates the relations hook **once per distinct relation** (deduped with the include-driven checks, via shared `RelationsSchema` instances) and prunes every dependent key across the whole `Query` — filter-tree included, with `elemMatch` prefix tracking.
- **Standalone safety.** Direct `parser.parseFilters(...)` / `codec.decodeFilters(...)` self-authorize, so per-parameter parses aren't a bypass either.
- **Clean surface.** The ledger is an explicit internal driver argument (`IQueryParameterParser.parseParameter(input, options, ledger)`), never part of the public parse options.

Covers all three filter dialects, including operators applied directly to a relation array (`$size` / `$all` / `$elemMatch`, expression `size`) and excluded dotted fields — both auto-join without materialising a column.

## Design decisions to sign off on

Two properties are **intentional** and worth a conscious decision, not just inheritance:

1. **Server-authored schema `default`s bypass the hook.** A `default: { 'user.name': 'DESC' }` materialises even for an actor denied `user`. This is consistent — the same default applies whether or not the client sends input, so it is not an actor *escalation* — and matches the documented "defaults are trusted" model. But it does mean "actor denied X, yet X is joined via a server default". If that is not desired, defaults would need to be actor-gated (a separate semantics change).
2. **Authorization is a property of the parse/decode boundary, not the `Query` AST.** A query assembled server-side with the schema-free build layer (`defineQuery`, condition helpers) and handed straight to an adapter is **not** re-checked. This is deliberate — authorization belongs at the trust boundary where untrusted input is parsed — but it means the guarantee is "`decode()`/`parse()` is safe", not "every `Query` object is safe".

Both are documented in `guide/recipes/authorization.md`.

## Rationale (please don't re-litigate)

- **Why the choke point (not per-parser recording).** Instrumenting each parser's resolution sites is fail-*open*: a new operator/dialect silently reopens the bypass (the review found exactly this — `$all`/`$size`). Recording in `resolveKey` makes coverage a structural invariant, tripwired by the operator matrices.
- **Why the ledger is an explicit driver arg, not an options field.** The earlier `relationValidations?` on the public parse options was an internal orchestration channel leaking into the API (and `stripInternal` forced it public across the package boundary). `parseParameter(input, options, ledger)` keeps it internal.
- **Why over-firing is acceptable.** Recording on resolve (before a leaf might be dropped) can, in a very narrow case, authorize a relation whose only reference was dropped — the fail-*safe* direction, deduped to one hook call, and a value-decode failure throws its own error first.
- **Why obligations evaluate under the relations schema's policy** (not the parameter's) — a relation join is a relations concern regardless of which parameter forced it; the expression dialect's throwing key-resolution still *drops* relation rejections unless the relations schema opts in.
- **Why no shared parser base class** — it would erase the per-call `<RECORD>` generic on `parse` (the typed key-path feature), so the per-parser entry methods stay.

## Commits

1. `feat(core)` — the fix (resolveKey recording, pooled ledger + prune, `parseParameter` driver) across core + the three filter dialects + docs.
2. `test(core)` — restore core coverage by unit-testing `relation-prune.ts` and the `resolveKey` obligation sink directly (the fix added code exercised only downstream); plus a doc-anchor fix.
3. `test` — harden: a direct `BaseQueryParser` orchestration test (pool/dedup/prune/throwOnFailure/async), cross-dialect operator matrices, and the async codec path.

## Testing

- New `relations-traversal` specs across parser-simple, parser-expression, parser-mongo and codec-url: include-less bypass closed, single-firing/dedup, deep-path cascade, `throwOnFailure`, async parity, standalone enforcement, `elemMatch`/`$size`/`$all` relation targets, and per-dialect operator-matrix tripwires.
- A direct `BaseQueryParser` test and unit tests for the prune helpers + resolver obligation sink in `@rapiq/core`.
- Full suite green (all packages), lint clean, core coverage restored (`query.ts` 0% → ~84%). Docs updated (`guide/schemas.md`, `guide/recipes/authorization.md`).


tada5hi:
@coderabbitai pause

PR #825, opened 2026-07-23
fix(typeorm): hydrate included relations & allow json columns in fields (#824)

### LINKED ISSUE #824: fields projection can't express json/array columns, and includes without joinAndSelect+fields are silently dropped
## Summary

Two related projection issues in `@rapiq/core` + `@rapiq/typeorm` (2.0.0-beta.7) that, together, make it impossible to have an entity that is **both** an `include=` target **and** returns its json/array columns. They surfaced as a production crash in a downstream app (see "Downstream impact") and an already-shipped silent data loss.

Self-contained repro at the bottom — 5 passing assertions in the `@rapiq/typeorm` test harness.

## Gotcha 1 — a relation included without `joinAndSelect` **and** without projected child fields is silently dropped

`TypeormAdapter` joins relations with `leftJoin` unless `relations.joinAndSelect` is set. When it is *not* set, a relation only hydrates if the child's columns are present in the `Fields` projection (they get selected via the root `qb.select(columns)`). If the query has a **root** `fields` projection (so `qb.select()` is active) but the included relation contributes **no** fields and `joinAndSelect` is off, the relation is joined but nothing of it is selected → TypeORM never hydrates it → the relation comes back `undefined`, with no error.

```ts
// root has a projection, child included with neither joinAndSelect nor child fields
new Query({
  fields: new Fields([new Field('id'), new Field('name')]),
  relations: new Relations([new Relation('child')]),
});
// -> parent.child === undefined   (G1a)

// setting joinAndSelect hydrates it (with all columns, json included)
new TypeormAdapter({ queryBuilder, relations: { joinAndSelect: true } });
// -> parent.child defined, parent.child.args === [...]   (G1b)
```

This is the footgun that bit two downstream services: whenever a schema is used as an `include` target but has no explicit `fields` block, the relation silently vanishes for every consumer that doesn't set `joinAndSelect`. It reads as an authorization/data bug and is hard to trace.

**Suggested direction:** when a relation is included but has no projected fields and `joinAndSelect` is off, still hydrate it — select the child's columns (or at least its PK) rather than dropping it. That makes `include` self-consistent regardless of whether the child schema declares `fields`.

## Gotcha 2 — json/array columns cannot be expressed in a `fields` projection (`SimpleKeys<T>`), so they are unavoidably dropped

A json column (typed as an array/object in TS but a single scalar column at the DB level, e.g. `@Column({ type: 'simple-json' }) args: {k:string}[]`) selects perfectly well when placed in the `Fields` IR (G2a). But `defineSchema<T>({ fields: { default: [...] } })` **rejects** the key at the type level, because `SimpleKeys<T>` excludes array/object-typed properties (G2c):

```ts
// @ts-expect-error 'data' is array-typed -> not a SimpleKey -> rejected
defineSchema<Parent>({ name: 'parent', fields: { default: ['id', 'name', 'data'] } });
```

Consequently a schema that uses an explicit `fields` projection can never include its json columns, and any explicit projection that omits them drops them at runtime (G2b). Combined with Gotcha 1 (an include-target schema *needs* explicit `fields` to hydrate under the no-`joinAndSelect` usage), an entity that is both an include target and has a json column is stuck: give it `fields` and the json is dropped; omit `fields` and the relation is dropped.

**Suggested direction:** allow column-backed json/`simple-json` columns to be listed in `fields` (widen the projectable-key type beyond `SimpleKeys` for scalar-at-DB columns), or provide an "always-select these columns" escape hatch on the schema.

## Downstream impact

- **authup** (`authup/authup#3313`): `include=permission` returned `undefined` on the client-/user-/role-permission collections because `permissionSchema` had no `fields` block (Gotcha 1). Worked around by adding a `fields` block (permission is all-scalar, so Gotcha 2 didn't bite). This crashed FLAME Hub node creation (`clientPermission.permission.name` on `undefined`).
- **FLAME Hub**: `include=master_image` drops the whole relation (Gotcha 1, `masterImageSchema` has no `fields`); and `GET /analyses` already silently drops `image_command_arguments` (Gotcha 2 — the analysis schema has explicit `fields` but its json column can't be listed). Hub can't use `joinAndSelect: true` as an escape because it paginates with `GROUP BY <root>.id` (postgres would reject the non-grouped joined-and-selected columns), so it depends on the fields-select path — which both gotchas break.

Fixing either gotcha in rapiq resolves the downstream cases without lossy per-schema workarounds.

## Reproduction (drop into `packages/typeorm/test/unit/`, all 5 pass)

```ts
import 'reflect-metadata';
import {
    Column, DataSource, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn,
} from 'typeorm';
import {
    Field, Fields, Query, Relation, Relations, defineSchema,
} from '@rapiq/core';
import { TypeormAdapter } from '../../src';

@Entity()
class Child {
    @PrimaryGeneratedColumn() id: number;
    @Column() name: string;
    @Column({ type: 'simple-json', nullable: true }) args: { k: string }[] | null;
}

@Entity()
class Parent {
    @PrimaryGeneratedColumn() id: number;
    @Column() name: string;
    @Column({ type: 'simple-json', nullable: true }) data: { k: string }[] | null;
    @Column({ nullable: true }) child_id: number | null;
    @ManyToOne(() => Child, { nullable: true }) @JoinColumn({ name: 'child_id' }) child: Child;
}

describe('repro projection', () => {
    let dataSource: DataSource;
    beforeAll(async () => {
        dataSource = new DataSource({ type: 'better-sqlite3', database: ':memory:', entities: [Parent, Child] });
        await dataSource.initialize();
        await dataSource.synchronize();
        const child = await dataSource.getRepository(Child).save({ name: 'c', args: [{ k: 'a' }] });
        await dataSource.getRepository(Parent).save({ name: 'p', data: [{ k: 'b' }], child_id: (child as any).id });
    });
    afterAll(async () => { await dataSource.destroy(); });

    const run = (query: Query, relationOpts: any = {}) => {
        const qb = dataSource.getRepository(Parent).createQueryBuilder('parent');
        new TypeormAdapter({ queryBuilder: qb, relations: relationOpts }).execute(query);
        return qb.getOneOrFail();
    };

    it('G1a: include child WITHOUT joinAndSelect and WITHOUT child fields -> child NOT hydrated', async () => {
        const parent = await run(new Query({
            fields: new Fields([new Field('id'), new Field('name')]),
            relations: new Relations([new Relation('child')]),
        }), {});
        expect(parent.child).toBeUndefined();
    });
    it('G1b: same but WITH joinAndSelect:true -> child hydrated WITH its json column', async () => {
        const parent = await run(new Query({
            fields: new Fields([new Field('id'), new Field('name')]),
            relations: new Relations([new Relation('child')]),
        }), { joinAndSelect: true });
        expect(parent.child).toBeDefined();
        expect(parent.child.args).toEqual([{ k: 'a' }]);
    });
    it('G2a: a json column selects fine when placed in the Fields IR', async () => {
        const parent = await run(new Query({ fields: new Fields([new Field('id'), new Field('data')]) }));
        expect(parent.data).toEqual([{ k: 'b' }]);
    });
    it('G2b: json column dropped by an explicit projection that omits it', async () => {
        const parent = await run(new Query({ fields: new Fields([new Field('id'), new Field('name')]) }));
        expect(parent.data).toBeUndefined();
    });
    it('G2c: defineSchema fields cannot list the json column (type-level)', () => {
        // @ts-expect-error 'data' is array-typed -> not a SimpleKey -> rejected by fields
        defineSchema<Parent>({ name: 'parent', fields: { default: ['id', 'name', 'data'] } });
        defineSchema<Parent>({ name: 'parent2', fields: { default: ['id', 'name'] } });
        expect(true).toBe(true);
    });
});
```

## Environment

`@rapiq/core` / `@rapiq/typeorm` 2.0.0-beta.7, typeorm 1.1, better-sqlite3 (also reproduces on the projection path independent of driver).

tada5hi:
Fixes #824.

Two related projection issues in `@rapiq/core` + `@rapiq/typeorm` that together made it impossible to have an entity be both an `include=` target and return its json columns.

## Gotcha 1 — an included relation without `joinAndSelect` + child fields was silently dropped

`TypeormAdapter` joined an `include`d relation with a plain `leftJoin` unless `relations.joinAndSelect` was set. Without a child-field projection, TypeORM never hydrated it, so the relation came back `undefined` — an authorization/data bug that was hard to trace (crashed downstream node creation on `clientPermission.permission.name`).

An included relation is now **join-and-selected as a whole subtree**, so it hydrates on its own. This deliberately matches the settled `@rapiq/memory` projection contract (`should widen refined relations to the whole subtree`):

- `include=child` → child fully hydrated, **even alongside** a sparse `fields=[…, child.name]` (a sparse child field never narrows an included relation).
- `fields=child.name` **without** an include → child materialized sparsely (join-for-projection only).
- a relation joined only for a **filter/sort** → plain `leftJoin`, never selected.
- `relations.joinAndSelect: true` → unchanged (selects every joined relation).

Relations now carry an `include` flag (set by the relations visitor, upgraded — never downgraded — when a field/filter/sort traversal also names them), and `shouldSelect = joinAndSelect || include`.

## Gotcha 2 — json/array columns could not be listed in a `fields` schema block

A concrete-typed json column (`@Column({ type: 'simple-json' }) args: { k: string }[]`) selects fine through the `Fields` IR but was rejected by `SimpleKeys<T>` at the `defineSchema` type level, so it could never appear in `fields.allowed`/`default` and was unavoidably dropped.

The `fields` key type is widened to `FieldKeys<T> = SimpleKeys<T> | SimpleResourceKeys<T>`, which admits concrete json shapes (index-signature json like `Record<string, any>` was already a `SimpleKeys`).

A json column and a relation are structurally identical to the type system, so relation keys are admitted here too. Rather than add a backend-specific guard, this follows the maintainer's steer: **`fields` is for columns, `relations` is for relations.** Correct schema-driven input never produces a relation key in `fields` (a client's `fields=child` is dropped by the `fields.allowed` allow-list at decode time, since `child` lives in `relations.allowed`); a hand-built `new Field('child')` is caller responsibility, not something a backend silently papers over.

## Notes / design

- No new cross-backend behavior: the change makes `@rapiq/typeorm` match `@rapiq/memory`, and `@rapiq/sql` is untouched (it never joins/hydrates).
- The `fields.execute()` → `relations.execute()` ordering is now documented as load-bearing (`select()` resets, join-and-select appends).
- Two independent adversarial reviews drove this design; both flagged an earlier `hasProjection`-based approach as over-hydrating on deeper projections and diverging from the memory contract — that machinery was removed in favor of the simpler `joinAndSelect || include` rule.

## Follow-up (out of scope)

The client-side build layer (`defineFields`/`defineQuery`, `FieldsBuildInput`) still types field keys via `SimpleKeys`/`NestedKeys`, so it rejects concrete json columns — an asymmetry with the (now widened) server schema. Worth a separate change.

## Tests

- `packages/typeorm/test/unit/hydration.spec.ts` (new): include-without-joinAndSelect, whole-subtree widening, sparse traversal, nested `child.pet` includes (incl. alongside a filter), filter-only-not-hydrated, json projection, and an end-to-end `SchemaRegistry` + `SimpleParser` → `TypeormAdapter` case mirroring the downstream crash.
- `packages/core/test/unit/types.spec.ts`: `FieldKeys` admits a concrete json column that `SimpleKeys` rejects.
- Docs: `packages/docs/packages/typeorm.md` `relations.joinAndSelect` description updated.

All 8 test projects pass; full build + lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting typed JSON and resource-shaped fields in field projections.
  * Included relations are now fully hydrated, including nested relations and JSON/array columns.
  * Relations used only for filtering or sorting remain unselected unless explicitly included.

* **Bug Fixes**
  * Improved relation hydration when `joinAndSelect` is disabled.

* **Documentation**
  * Clarified relation join and hydration behavior for the TypeORM adapter.

### DIFF
--- a/packages/core/src/schema/parameter/fields/schema.ts
+++ b/packages/core/src/schema/parameter/fields/schema.ts
@@ -5,7 +5,7 @@
  * view the LICENSE file that was distributed with this source code.
  */
 
-import type { MaybeAsync, ObjectLiteral, SimpleKeys } from '../../../types';
+import type { FieldKeys, MaybeAsync, ObjectLiteral } from '../../../types';
 import {
     isPropertyNameValid,
 } from '../../../utils';
@@ -63,7 +63,7 @@ export class FieldsSchema<
 
     // ---------------------------------------------------------
 
-    setDefault(input?: SimpleKeys<RECORD>[]) {
+    setDefault(input?: FieldKeys<RECORD>[]) {
         if (typeof input === 'undefined') {
             this.default = [];
             this.defaultIsUndefined = true;
@@ -74,7 +74,7 @@ export class FieldsSchema<
         this.defaultIsUndefined = false;
     }
 
-    setAllowed(input?: SimpleKeys<RECORD>[]) {
+    setAllowed(input?: FieldKeys<RECORD>[]) {
         if (typeof input === 'undefined') {
             this.allowed = [];
             this.allowedIsUndefined = true;
--- a/packages/core/src/schema/parameter/fields/types.ts
+++ b/packages/core/src/schema/parameter/fields/types.ts
@@ -6,15 +6,15 @@
  */
 
 import type { BaseSchemaOptions, KeyValidator } from '../../types';
-import type { SimpleKeys } from '../../../types';
+import type { FieldKeys } from '../../../types';
 
 export type FieldsOptions<
     T extends Record<string, any> = Record<string, any>,
     CONTEXT = any,
 > = BaseSchemaOptions & {
     mapping?: Record<string, string>,
-    allowed?: SimpleKeys<T>[],
-    default?: SimpleKeys<T>[],
+    allowed?: FieldKeys<T>[],
+    default?: FieldKeys<T>[],
     /**
      * Dynamic per-field gate, e.g. an actor permission check.
      * Runs once per client-requested field against the schema that
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -118,6 +118,22 @@ export type SimpleResourceKeys<
             never
 }[keyof T & string];
 
+/**
+ * Keys a `fields` projection may list: scalar/leaf columns ({@link SimpleKeys})
+ * plus record/array-shaped keys ({@link SimpleResourceKeys}). The latter covers
+ * concrete-typed json columns (e.g. `{ k: string }[]`) — single columns at the
+ * database level that {@link SimpleKeys} would otherwise reject. Index-signature
+ * json (`Record<string, any>`) is already a {@link SimpleKeys}.
+ *
+ * A json column and a relation are structurally identical at the type level, so
+ * relation keys are admitted here too. List only column-backed keys in `fields`;
+ * relations belong in the `relations` allow-list (`include`). A relation key that
+ * reaches a backend's `fields` is caller error — it is not a selectable column.
+ */
+export type FieldKeys<
+    T extends Record<PropertyKey, any>,
+> = SimpleKeys<T> | SimpleResourceKeys<T>;
+
 export type NestedResourceKeys<
     T extends Record<PropertyKey, any>,
     DEPTH extends number = 4,
--- a/packages/core/test/unit/types.spec.ts
+++ b/packages/core/test/unit/types.spec.ts
@@ -8,6 +8,7 @@
 import { expectTypeOf } from 'vitest';
 import { defineSchema } from '../../src';
 import type {
+    FieldKeys,
     NestedKeys,
     NestedResourceKeys,
     SimpleKeys,
@@ -62,6 +63,30 @@ describe('src/types.ts', () => {
 
             expect(keys).toBeDefined();
             expect(invalid).toBeDefined();
+        });
+    });
+
+    describe('FieldKeys', () => {
+        it('should admit a concrete-typed json column that SimpleKeys rejects (#824)', () => {
+            type Row = {
+                id: string,
+                name: string,
+                args: { k: string }[] | null,
+            };
+
+            const keys: FieldKeys<Row>[] = ['id', 'name', 'args'];
+
+            // @ts-expect-error a concrete-shaped json column is not a SimpleKey
+            const rejected: SimpleKeys<Row>[] = ['args'];
+
+            expect(keys).toBeDefined();
+            expect(rejected).toBeDefined();
+        });
+
+        it('should union leaf columns and resource-shaped keys', () => {
+            const keys: FieldKeys<Event>[] = ['id', 'data', 'meta', 'created_at', 'realm', 'user', 'items'];
+
+            expect(keys).toBeDefined();
         });
     });
 
--- a/packages/docs/packages/typeorm.md
+++ b/packages/docs/packages/typeorm.md
@@ -59,7 +59,7 @@ new TypeormAdapter({
 
 | Option | Description |
 |---|---|
-| `relations.joinAndSelect` | Hydrate **every** joined relation, including those joined only for a filter or sort. Off by default — an `include`d relation is still hydrated on its own (join-and-selected, or via its columns when a `fields` projection already covers it), while a relation joined purely for filtering/sorting stays unselected. |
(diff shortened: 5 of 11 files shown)


tada5hi:
@coderabbitai pause

PR #832, opened 2026-07-24
feat(typeorm): id-only relation hydration for grouped pagination (#828)

### LINKED ISSUE #828: Document (and consider an escape for) include hydration vs GROUP BY <root>.id on strict dialects
## Summary

After #825, an explicitly `include`d relation is always `leftJoinAndSelect`ed as a **complete subtree** (matching the `@rapiq/memory` projection contract). That is the intended semantics, but it interacts badly with the `onJoin` + `GROUP BY <root>.id` pagination pattern on strict SQL dialects (postgres): the join selects *every* child column, and postgres rejects non-grouped selected columns under `GROUP BY`.

This is not a regression in behavior that ever worked (before #825 the relation came back `undefined`), and it already applies to `relations.joinAndSelect: true`. But it deserves explicit documentation — and possibly a first-class escape — because it directly affects the downstream case #824 set out to fix.

## Why it happens

`TypeormAdapter` hydrates an included relation with `leftJoinAndSelect(parent.child, alias)`, which adds **all** of the child's columns to the `SELECT`. A consumer paginating with `GROUP BY parent.id` (e.g. via `relations.onJoin: (_, alias, qb) => qb.addGroupBy(`${alias}.id`)`) then produces:

```sql
SELECT parent.id, ..., child.id, child.name, child.args, ...
FROM parent
LEFT JOIN child ON ...
GROUP BY parent.id, child.id
```

Postgres rejects `child.name`, `child.args`, … — they are neither grouped nor aggregated. Hydrating arbitrary joined columns and collapsing to distinct roots are fundamentally incompatible in strict SQL, regardless of how the columns get selected.

## Downstream relevance

FLAME Hub paginates with `GROUP BY <root>.id` and relies on the fields-select path rather than `joinAndSelect` (postgres would otherwise reject the joined-and-selected columns — see #824). With #825, `include=<relation>` on such a query full-hydrates the relation and hits exactly this conflict, so the fix does not "just work" for the grouped-pagination path.

## Current workaround (works today)

Use **field traversal** instead of `include` for grouped/sparse selection: `fields=child.col` joins the relation with a plain `leftJoin` and selects only the referenced (groupable) column — no `leftJoinAndSelect`. `include` remains the right tool when you are *not* grouping the root.

## Suggested direction

1. **Document the interaction** (minimum) — a note near the `relations.onJoin` / `joinAndSelect` docs: an `include`d relation is join-and-selected as a whole subtree, which cannot coexist with `GROUP BY <root>.id` on strict dialects; use `fields=<relation>.<column>` for grouped sparse selection, or a non-grouping pagination strategy (distinct-root subquery / two-phase id-then-load) when full includes are needed.
2. **(Optional) consider an escape** — e.g. a `relations` hydration mode that selects only groupable columns (the PK) for includes so the relation object is at least defined under `GROUP BY`, acknowledging it cannot be a full subtree. Limited value (id-only), so this is a design question, not a clear win.

## References

- #824 (parent), #825 (include hydration)

tada5hi:
Closes #828.

## Context

After #825, an explicitly `include`d relation is `leftJoinAndSelect`ed as a **complete subtree** (matching the `@rapiq/memory` projection contract). That is the intended semantics, but it cannot coexist with the `GROUP BY <root>.id` pagination pattern on strict dialects (postgres): the join selects *every* child column, and postgres rejects non-grouped selected columns under `GROUP BY`. The same already applies to `relations.joinAndSelect: true`.

This PR does both parts of #828: **documents the interaction** (the minimum) and adds a **first-class escape**.

## The escape — `relations.hydrationMode: 'full' | 'key'`

New adapter option controlling the select granularity of the relations the adapter hydrates (parallels the existing `joinType`):

- `'full'` (default) — the whole subtree (`leftJoinAndSelect`); unchanged behavior.
- `'key'` — a plain `leftJoin` + `addSelect(<alias>.<pk>)` per primary column, so the relation object is hydrated **id-only** and every selected join column is groupable under `GROUP BY <root>.id`.

```ts
new TypeormAdapter({
    queryBuilder,
    relations: {
        hydrationMode: 'key',
        onJoin: (_p, alias, qb) => qb.addGroupBy(`${alias}.id`),
    },
});
```

### Why id-only actually hydrates

Verified against the TypeORM 1.1.0 transformer: `JoinAttribute.isSelected` returns true when any selected column matches `<alias>.<propertyPath>`, so selecting only the primary key is enough for the result transformer to build the nested `{ id }` object. A plain `leftJoin` with no select would leave the relation `undefined`.

### Duplicate-select guard

`relations.execute()` runs after `fields.execute()`, so a `<relation>.<pk>` field path — or a manual caller `.select()`/`.addSelect()` on the relation alias — may already cover the key. The key-mode add reads the live `expressionMap.selects` and skips a primary key that is already selected (by rapiq or externally), so the adapter never stacks duplicate SELECT columns. (Note: TypeORM itself emits a manually-selected relation PK twice via its two select-generation passes — a pre-existing quirk the guard does not amplify.)

## Docs

`packages/docs/packages/typeorm.md`:
- New `relations.hydrationMode` options-table row + example.
- A warning block explaining the `include` + `GROUP BY <root>.id` conflict and the three escapes: sparse `fields=<relation>.<column>`, `hydrationMode: 'key'`, or a non-grouping pagination strategy (distinct-root subquery / ids-then-load).

## Limitation (documented, by design)

`'key'` hydrates the relation object id-only — the subtree is not materialized. It is the right tool when a caller relies on the relation being *present* but does not need its columns; when full columns are needed under a grouped root, use a non-grouping pagination strategy instead.

## Tests

`packages/typeorm/test/unit/hydration.spec.ts` (against a real in-memory sqlite builder):
- id-only hydration (child defined, only `id` populated);
- SQL shape: plain `LEFT JOIN` selecting only the key;
- nested id-only include;
- id-only include staying defined under `GROUP BY <root>.id` + `onJoin`;
- `hydrationMode: 'full'` matching the default;
- duplicate-select guard (key not re-added when a field already names it).

All 155 `@rapiq/typeorm` tests pass; build + lint clean.

### DIFF
--- a/packages/docs/packages/typeorm.md
+++ b/packages/docs/packages/typeorm.md
@@ -49,6 +49,7 @@ adapter.execute(query, { clear: false });
 new TypeormAdapter({
     relations: {
         joinAndSelect: true,
+        hydrationMode: 'full',
         joinType: 'left',
         onJoin: (path, alias, queryBuilder) => {
             queryBuilder.addGroupBy(`${alias}.id`);
@@ -60,6 +61,7 @@ new TypeormAdapter({
 | Option | Description |
 |---|---|
 | `relations.joinAndSelect` | Hydrate **every** joined relation, including those joined only for a filter or sort. Off by default — an `include`d relation is always hydrated as a complete subtree (a sparse `fields` entry such as `child.name` never narrows it), while a relation joined purely for filtering/sorting stays unselected (its columns are hydrated only when a `fields` path references them). |
+| `relations.hydrationMode` | `'full'` (default) or `'key'`. Select granularity for the relations the adapter hydrates. `'full'` selects the whole subtree (`leftJoinAndSelect`); `'key'` selects only the relation's primary key (plain `leftJoin` + `addSelect(<alias>.<pk>)`), so the relation object is hydrated **id-only**. Use `'key'` to keep an `include`d relation defined under `GROUP BY <root>.id` on strict dialects — see the warning below. |
 | `relations.joinType` | `'left'` (default) or `'inner'`. Left joins keep records whose relation is absent. |
 | `relations.onJoin` | Invoked as `(path, alias, queryBuilder)` for every join the adapter applies — e.g. to `addGroupBy` per join when the root query is grouped. Skipped (pre-existing) joins don't trigger it. |
 | `relations.relationAlias` | Derive the join alias for a relation path (default: collision-free length-prefixed segments, e.g. `role.realm` → `r4_role_5_realm`). Filter/sort/field references resolve against the same derivation. |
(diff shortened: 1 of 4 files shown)


tada5hi:
@coderabbitai pause

Issue #810, opened 2026-07-21
RFC: relation-scoped conditions (row-level narrowing of included relations)


tada5hi:
## Problem

The context-threading RFC (#806) made include authorization **all-or-nothing per relation**: `relations.validate` either admits `include=client` or drops it. Row-level narrowing of the *included* rows is not expressible in the IR:

- *"You may include `client` — but only client rows in your realm."*
- *"Joined rows the actor may not read come back `null`; the parent rows are unaffected."*

Today the only options are dropping the include entirely (coarse), or post-filtering loaded rows in the application (per-row permission checks, N+1, and hand-rolled redaction loops). Neither expresses the common authorization outcome: the parent collection is untouched while the included payload is scoped to the actor.

Note the contrast with what the IR *can* already express: a filter on a dotted path (`filters['client.realm_id']`) constrains the **parent** row set. This RFC is about the other axis — constraining which **related** rows materialize, without touching parent selection, pagination or totals.

### 1. IR: conditions bound to a relation

`Relation` gains an optional condition (working shape — bikeshed in open questions):

```ts
class Relation {
    readonly name: string;
    readonly condition?: ICondition;   // constrains the *related* rows, never the parent
}
```

Conditions are ordinary `Filters`/`Filter` trees whose field names are relative to the related record (`eq('realm_id', actorRealmId)` on the `client` relation). Deep paths compose per level: a condition on `client` and another on `client.realm` each scope their own node.

### 2. Semantics per backend

| Backend | Rendering | Effect |
|---|---|---|
| `@rapiq/sql` / `@rapiq/typeorm` | condition ANDed into the **join ON clause** (`LEFT JOIN client c ON ... AND <condition over c>`), placeholders via the existing dialect machinery, alias via the shared `buildRelationAlias` | non-matching joined rows are `NULL`; parent row count, pagination and totals unchanged |
| `@rapiq/memory` | sub-object predicate compiled with the existing `compileFilters` machinery, applied when projecting the included relation | to-one: non-matching object → `null`; to-many: array filtered; keep-tree projection retains the key |

Two invariants:

- **Parent invariance.** A relation-scoped condition never removes parent rows. (Filtering parents by related rows remains the job of dotted *filters* — unchanged.)
- **SQL/memory parity.** Same condition tree, same visible result — extending the existing parity contract.

[...]

PR #837, opened 2026-07-25
feat: condition-returning and batched key validation hooks (#830)

### LINKED ISSUE #830: Feature: location-aware KeyValidator (root-only fields — allow at query root, strip via include)
## Summary

Thread the **resolution location** (root vs. via-relation, plus the traversed
relation path) into `KeyValidator` so a `fields.validate` / `sort.validate` /
`relations.validate` hook can gate a key based on *where it sits in the query
tree*, not only on its canonical name.

The concrete need: express a **root-only field** — a column that may be
projected when its schema is the query root, but must be stripped when the same
schema is reached as an **include target** on another entity.

## Motivation (real use case)

In [authup](https://github.com/authup/authup) the `client` schema has a
`secret` column. It is legitimately selectable on the client's own endpoint:

```
GET /clients/:id?fields=+secret
```

because that request is served by `ClientService`, which applies a per-row
authorization gate (plaintext secrets are only returned to an actor whose
permission covers that specific row/realm).

The same `client` schema is also an **include target** of junction entities
(`client-permission`, `client-role`, `client-scope`) which all allow
`include=client`. Via the include path the service gate never runs:

```
GET /client-permissions?include=client&fields[client]=secret
```

is served by `ClientPermissionService`, so the client's per-row secret gate is
bypassed and the secret is projected. `relations.validate` (the #815 gate) only
requires the *coarse* "may read clients at all" pre-check to permit the join, so
a realm-scoped reader who passes that pre-check can read another realm's client
secret through the include.

The natural home for the guard is the **client schema itself** (so it applies
everywhere the schema governs a projection). But today the schema layer cannot
express "allow `secret` at the root, strip it when reached via a relation":

- The effective allow-list is `allowed ∪ default` (`FieldsSchema.isValid`), so
  putting `secret` in `allowed` makes it projectable in **both** positions.
- `fields.validate` fires for the include-child key, but the hook is
  **path-blind**. `applyKeySchemaValidation[Async]` calls:

  ```ts
  entry.schema.validate(entry.key, context)
  ```

  and `KeyValidator = (name, context) => MaybeAsync<boolean | undefined>`. The
  hook receives only the canonical key (`"secret"`) — identical for root
  `fields=secret` and include-child `fields[client]=secret`. `PendingKeyValidation`
  already carries `path` (the dotted output name, e.g. `client.secret`) and
  `key`, but only `key` reaches the hook, so the hook can't branch on location.

The only safe workaround today is to drop `secret` from the allow-list
entirely, which also kills the legitimate root `fields=+secret` read. That's a
functional regression used purely to close a security hole — exactly the kind of
gap worth fixing upstream.

## Proposal

Make the resolution location available to `KeyValidator`. It already exists on
the recorded `PendingKeyValidation` (`key` + `path`); it just isn't handed to
the hook. Some options (any is fine):

1. **Richer hook argument** — pass a read-only descriptor alongside the context,
   e.g.

   ```ts
   type KeyValidatorLocation = {
     key: string;        // canonical key relative to the governing schema ("secret")
     path: string;       // full dotted output path ("client.secret")
     segments: string[]; // canonical relation segments traversed (["client"]) — [] at root
     isRoot: boolean;    // segments.length === 0
   };

   type KeyValidator<CONTEXT = any> = (
     name: string,
     context: CONTEXT,
     location: KeyValidatorLocation,   // additive, backward compatible
   ) => MaybeAsync<boolean | undefined>;
   ```

2. **Fold the location into a wrapper context** the resolver builds per
   obligation, if you'd rather not widen the arity.

Applies uniformly to all three `KeyValidatableSchema` implementers
(`FieldsSchema`, `SortSchema`, `RelationsSchema`) since they share
`applyKeySchemaValidation`. Existing `(name, context)` hooks keep working — the
location is purely additive.

With this, authup could write:

```ts
fields: {
  default: [/* … */],
  allowed: ['secret'],
  validate: (name, ctx, loc) => {
    if (name === 'secret' && !loc.isRoot) return false; // never via an include
    return gateSecretForActor(ctx.actor);               // root: normal gate
  },
}
```

so the sensitive field is root-only, stripped on every include path, with no
loss of the root read.

## Generalization

Any schema shared between a root query and an include projection that wants
**context-dependent field exposure** benefits — secret/credential columns, PII
columns (`email`, tokens), internal-only fields. "Selectable as a root, hidden
as a relation" is a recurring authorization shape once one schema serves both
roles.

## Prior art / related

- #806 — request-context threading & per-parameter validate hooks (the
  mechanism this extends).
- #815 — `relations.validate` now also prunes dependent filter/field/sort keys
  that traverse a denied relation; this proposal is the field-level analogue
  (gate a *field* by the relation it's reached through, without denying the
  whole relation).
- #810 — relation-scoped conditions (row-level narrowing) is adjacent but
  orthogonal: that narrows *rows* of an included relation; this gates *which
  columns* of an included relation may be projected.

tada5hi:
Closes #830.

## The issue changed shape twice

#830 was filed as *"allow a field at the query root, strip it via an include"*. Both halves of that framing were superseded:

1. The follow-up comment on the issue already called the coarse boolean a stopgap and pointed at a condition-producing field authorization as the end state.
2. During implementation the **root-only premise was dropped**: `secret` has to stay includable on `/client-permissions?include=client&fields[client]=secret`. Strip-on-include is not the fix.

So the settled law is neither "root only" nor row narrowing:

> A `validate` / `validateMany` hook may answer with an `ICondition`, meaning **this column is visible only on rows satisfying C**. It never changes which rows come back, at the query root or under any relation path.

A pleasant consequence: **#810 is not a dependency**. Relation-scoped row narrowing stays entirely with that RFC, and nothing here pre-empts it.

## What lands

```ts
type KeyValidationScope   = { parameter, path, schema? };   // path: '' at the query root
type KeyValidationVerdict = boolean | ICondition | undefined;

type KeyValidator<C>     = (name, context: C, scope) => MaybeAsync<KeyValidationVerdict>;
type KeyValidatorMany<C> = (names: string[], context: C, scope) => MaybeAsync<Record<string, KeyValidationVerdict>>;
```

- **Location awareness**, the original ask, as a third argument on both hooks. `path` distinguishes the positions one schema occupies in a single query, so a hook can still branch on where a key sits, with a plain boolean and no condition involved.
- **`validateMany`**, invoked once per `(governing schema, relation path)` with that position's deduplicated keys, so an authorization policy compiles once per query instead of once per key. Mutually exclusive with `validate`; declaring both throws `SCHEMA_KEY_VALIDATOR_CONFLICT` at schema construction rather than on a request. A key absent from the returned record is **rejected**, read through `hasOwnProperty` so prototype members cannot forge an acceptance.
- **`Field.condition`**, carrying a gate through the IR. `sort` and `relations` have no column to gate, so a condition there is refused rather than silently ignored.

## Enforcement is split, because TypeORM forces it

Per-column masking (`case when c then col end`) is not implementable: `packages/typeorm/src/adapter/fields.ts` must emit bare `alias.property` selections because TypeORM resolves them against its alias map to hydrate entities. So:

| Backend | Behaviour |
|---|---|
| `@rapiq/memory` | Honours the gate **while projecting**; the property is dropped from records that fail it. |
| `@rapiq/sql`, `@rapiq/typeorm` | Project the column **unconditionally**; the gate is applied **after the fetch** via `applyFieldConditions`. |

This is **fail-open** on the SQL backends and is documented as such, with the post-fetch helper as the remedy. Two structural aids narrow the risk: the SQL fields machinery **force-projects every column a gate reads** (so the post-fetch pass always has its operands, even under a sparse fieldset; a missing operand would over-redact eq-style gates and let negated gates disclose), and `hasFieldConditions(query)` from core lets a response path assert that no gated column ships unredacted. It also amends the memory/SQL parity contract, recorded in `.agents/architecture.md`. The typeorm spec exercises the whole loop against a live database: the column comes back unredacted, the gate does not become a row filter, and the helper redacts exactly the failing rows without dropping one.

`@rapiq/memory` is declared an optional peer of `@rapiq/sql` and `@rapiq/typeorm`, since enforcing a gate otherwise needs a package a consumer has no other reason to install.

## Encoding

A gate is server-side state with no wire form, so `encode()` throws `FEATURE_UNSUPPORTED` (`fields:condition`) rather than emitting the bare field name, which would hand the next hop an ungated column. A client can never author a condition, so this is unreachable from the client encode path; it fires only when re-encoding a query you decoded server-side, for example to build pagination links. The schema-aware encode pass strips the conditions its own validation round trip derives, so it does not false-positive.

## Backward compatibility

Existing `(name, context) => boolean` hooks compile and behave identically; the third argument is additive and the widened return type is a superset. `applyKeySchemaValidation` keeps its `string[]` return.

## Hardenings from the pre-merge audit

An adversarial audit of the branch surfaced five reproduced defects; all are fixed and pinned by tests:

- **Rejection could widen the projection** (critical). A hook rejecting the only requested field emptied the fields node, which every backend reads as project-everything. Validation that empties the selection now falls back to the input-less projection minus the rejected names, so a denial neither widens nor resurrects.
- **Partial-row gate evaluation** (major). Nothing projected the columns a condition reads, so sparse fieldsets broke both directions (over-redaction and disclosure). Fixed by the operand force-projection above.
- **Redactor resurrection** (major). A deeper gate redacting under a parent re-disclosed a property the parent's own gate had removed; presence is now checked on the output. Redacted rows also keep their class prototype.
- **Per-obligation failure policy** (major). A child schema's `throwOnFailure: true` now governs its own validate rejections in pooled pending lists, matching its allow-list behaviour.
- **Gatedness detectability** (major). `hasFieldConditions` makes the fail-open contract assertable instead of convention-only.

Known and accepted: a boolean deny of an included relation's column is weaker than a condition on typeorm (full-hydrated includes fetch the column regardless); the docs steer sensitive include-reachable columns to condition verdicts. The pre-existing empty-node backend contract (reachable on master via exclude-all) stays a follow-up.

## Two judgement calls worth a second opinion

- **`Fields.merge()`** keeps its documented keyed-by-name, left-priority replace, so a gated field merged under an ungated one of the same name loses its gate. Not special-cased, because conditions only arise server-side after a parse and diverging from the documented merge law seemed worse than the exposure. Happy to reverse.
- **Declaring `@rapiq/memory` an optional peer** of the two SQL packages adds a dependency edge that did not exist. It is legal in the layer graph (memory is layer 1, typeorm layer 2) and optional, but it is a structural statement rather than a pure fix.

## Not in scope

- **#810** relation-scoped row narrowing, now genuinely independent.
- **Per-column masking**, blocked by entity hydration as above.
- **The filters `Validator`** still returns `IFilter | undefined`; widening it to `ICondition` is a different hook with a different vocabulary.
- **Feeding `validateMany` the effective projection** (post-defaults). Schema defaults are server-authored and keep bypassing the hooks, so `names` is the requested key set.

## Consumer follow-up

authup is the motivating consumer and is untouched here. The `attributeNames` thread from the issue discussion is filed as authup/authup#3321: synthesizing a record at the call site buys nothing (`flattenObject` is the identity on boolean leaves) and widening the shared `ATTRIBUTES` policy-data key would make realm-match and the row-value ABAC evaluator settle against fabricated row data, so that policy needs its own key behind a config discriminator.

## Summary by CodeRabbit

* **New Features**
  * Added row-scoped field visibility gates, allowing selected column values to be hidden for records that fail a condition.
  * Added batched `validateMany` hooks for fields, relations, and sorting, with scope-aware validation.
  * SQL-based results now support post-fetch field redaction; in-memory projections apply gates directly.
* **Bug Fixes**
  * Preserved visibility conditions through field resolution, relation pruning, and nested projections.
* **Documentation**
  * Expanded guidance on field authorization, condition behavior, validation errors, and wire-format limitations.
* **Errors**
  * Added clearer errors for conflicting validation hooks and unsupported conditioned-field URL encoding.


tada5hi:
@coderabbitai full review

PR #865, opened 2026-08-02
fix(adapters): cross-backend drift fixes from the architecture audit

tada5hi:
Follow-up to the 2026-08-02 architecture audit: the cross-backend drift it confirmed, fixed one commit per concern, each with the regression spec that would have caught it.

## Fixes

- **adapter-memory: `limit: 0` is a value, not absence.** The pagination slicer returned every row for `limit: 0` while typeorm/prisma/drizzle (engine-measured) return none, and drizzle's impossible-root encoding (`limit: 0`) leans on that reading. The slice now applies for any non-negative numeric limit; negative values stay absence. A drizzle engine spec pins the agreement with `applyQuery`.
- **adapter-prisma: drizzle hardening backported.** Own-property provider lookups (inherited object members answer unknown instead of posing as presets), the `distributeNegation` invariant guards (typed failure instead of silently emitting the inverted positive form), and the typed ITSELF gate (previously an emitted `$this` key leaning on prisma's unknown-field error). Deliberately not ported because prisma is immune by construction: the metadata prototype hole (array-based walker) and numeric sort-name rejection (array-form orderBy).
- **adapter-prisma: `findMany()` no longer fails open on field visibility gates (#830).** The runner refuses typed when the query's fields carry conditions, directing to `execute()` plus `applyFieldConditions` of @rapiq/adapter-memory; `count()` is unaffected. Documented in the Running the query section.
- **adapter-sql/typeorm: `caseSensitive` widened to `string[] | boolean`.** Core and the other three backends already accept the blanket opt-out (`true`); the SQL pair silently did not. Passes through to `planCondition` unchanged. The option key split (`visitor` vs `filters`) is untouched; that is an API-freeze question for the GA checklist.

## Tests

- Engine parity matrices aligned: the rows prisma had and drizzle lacked are ported (ordering complements, to-one null column, to-many presence, elemMatch null interior), and both packages gain the operator-enrollment tripwire the typeorm suite established. The tripwire immediately caught `gt`/`lte` enrolled in neither matrix; both backends gain rows, verified against the live engines (sqlite locally; postgres via the tests-db matrix).
- Prisma's inline engine condition list moved to `test/data/matrix.ts` so the tripwire runs in the default suite while the engine spec replays the same fixture.

## Docs

- Architecture guide records the audit rationales: the preset resolution invariant (user-supplied names throw typed everywhere; derived facts fall back documented), adapter-sql's protected members as an intended extension surface, and `ICondition` via `planCondition` as the filters contract of record (per-operator visitor methods are a legacy fast-path).

No public API breaks beyond the additive `caseSensitive` widening.

## Summary by CodeRabbit

- **New Features**
  - Added a global `caseSensitive: true` option for exact equality comparisons across supported SQL and TypeORM adapters.
  - Added documentation for case-sensitive filtering and safe handling of visibility-restricted Prisma queries.

- **Bug Fixes**
  - Pagination with an explicit limit of `0` now correctly returns no rows.
  - Prisma rejects unsupported or unsafe field-condition queries instead of returning unfiltered results.
  - Provider resolution no longer accepts inherited object property names as valid providers.

- **Tests**
  - Expanded coverage for filtering, relationships, null handling, negation, collection matching, and adapter parity.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -60,6 +60,8 @@ A `Schema<RECORD>` declares what a client *may* request per parameter (`allowed`
 
 `@rapiq/adapter-sql` is database-agnostic; per-database behavior is injected via `DialectOptions` callbacks (`escapeField`, `paramPlaceholder`, `regexp`). Presets live in `packages/adapter-sql/src/dialect/`. Regex strings pass through unchanged for the database engine to interpret and validate; JavaScript `RegExp` values contribute their `source` and `ignoreCase` flag.
 
+Preset resolution follows one fleet-wide invariant (audited 2026-08-02, do not re-flag): **user-supplied names throw typed everywhere; derived facts fall back documented**. prisma/drizzle validate a user-supplied provider name (a typo must throw, it would break every case-insensitive filter), while typeorm's `resolveQueryDialect` derives the dialect from the query builder's driver — the user supplied nothing, an exotic driver must still work, and the pg last-resort default is documented.
+
 ## Key Abstractions
 
 ### Query AST (core)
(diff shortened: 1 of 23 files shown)


tada5hi:
@coderabbitai pause

PR #871, opened 2026-08-02
fix: close the pre-GA audit gaps in code and docs

tada5hi:
## What

Everything surfaced by the pre-release docs audit, in three layers.

### Code

- **core**: `Filters.and()`/`or()` on an empty receiver returned the injected conditions as a flat root-AND, so a later `merge()` could silently displace a scoping condition injected onto a query that carried no filters. The injected conditions now land in a nested group; the wrap-and-inject guarantee is unconditional and a hostile replace-merge throws `FILTERS_NOT_FLAT`. Injecting nothing returns the receiver.
- **adapter-prisma / adapter-drizzle**: finish the #868 move; `caseSensitive` is now genuinely top-level on the constructor and execute options of both serializer adapters (the #868 commit assumed it already was, but both nested it under a `filters` wrapper). The docs claim and the code now agree.
- **core**: `PaginationParseError.limitExceeded` carried the default `NONE` code; it now throws `ErrorCode.LIMIT_EXCEEDED`.

### Docs: audit fixes

- README build sample used invalid filter input (`gte(18)` as a field value); the execute row now names the memory adapter.
- Landing surfaces (hero, feature grid, code tabs, pipeline figure, site description) cover all five adapters and three parser dialects, teach `defineQuery` and show the stamped expression wire v2 actually writes.
- `guide/filters.md` states operator support honestly per backend (prisma/drizzle raise typed errors for regex/mod/size/ITSELF instead of approximating); the expression-keyword claim about negated helpers is corrected.
- `adapter-typeorm` page drops the removed `visitor` execute wrapper; the prisma/drizzle pages document the real `caseSensitive` surface plus the schema-to-adapter forwarding step.
- relations/sort document their `validate`/`validateMany` hooks (including the traversal gating); fields and codec-url document the `$root` wire token; parser-simple documents the `context` option and the filter wire-grammar exports; core documents the plan layer; schemas regenerates the `describe()` example from its own schema; errors gains the missing `OPERATOR_UNSUPPORTED`/`LIMIT_EXCEEDED` rows.
- Broken heading anchors repaired, `integrations/url.md` redirects like its siblings, the docs workspace README is excluded from the site build, and every em/en dash is gone from the published prose.

### Docs: recipes as one story

[...]

PR #876, opened 2026-08-03
feat(core)!: carry filter displaceability explicitly

### LINKED ISSUE #875: Filters: carry displaceability explicitly instead of via tree nesting (make merge total)
## Context

Found while upgrading authup to `@rapiq/core` 2.0.0-beta.15, i.e. after the #871 empty-receiver wrap fix. That fix is correct and closes a real hole. This issue is about the *mechanism* it relies on, which seems worth settling before 2.0.0 freezes the API.

## The mechanism today

`and()` / `or()` wrap-and-inject produce a non-flat tree. `merge()` requires a structurally flat root-AND (`extractFlatConditions`) and throws `FILTERS_NOT_FLAT` otherwise. So the nesting is what makes an injected condition non-displaceable.

That means tree **structure** carries two unrelated meanings at once:

1. logical grouping
2. "this condition is protected from displacement"

## Problem 1: any normalization silently strips the protection

`flatten()` collapses same-operator nests, so it turns a protected tree back into a displaceable one:

```js
const scope   = eq('realmId', 'SCOPE');
const wrapped = new Query({}).filters.and(scope);
const user    = defineFilters(eq('realmId', 'ATTACKER'));

wrapped.value        // [{ operator: 'and', value: [eq(realmId,'SCOPE')] }]
user.merge(wrapped)  // throws filtersNotFlat                  <- protected

const flat = wrapped.flatten();
flat.value           // [eq(realmId,'SCOPE')]
user.merge(flat)     // [eq(realmId,'ATTACKER')]               <- scope silently displaced
```

So the #871 guarantee holds only as long as no caller flattens. Any present or future normalization pass re-opens exactly the hole #871 closed, silently. The intent is encoded in a representation detail rather than in the data.

## Problem 2: merge is partial in a way that conflates two cases

Two different situations raise the same error:

- **(a) genuinely compound trees**, e.g. `and(a, or(b,c)).merge(eq(b,9))`. Per-field replace has no sound meaning here: dropping the incoming `b` changes the disjunction non-locally, keeping it contradicts receiver priority. Refusing is right.
- **(b) wrap-injected trees**, which are often semantically a plain conjunction. These throw only because structure is being used as the protection marker.

Worth noting the build helpers already normalize, so `defineFilters(and(eq('a','1')))` is flat and merges fine. Only the `Filters.and()` / `Filters.or()` methods produce the nested shape.

## Proposal

Carry displaceability **explicitly** (a flag on the condition/group, or a distinct node kind) instead of inferring it from nesting. Then define `merge` over an AND-rooted tree as:

- root-level **leaf** conditions participate in per-field replace, exactly as today
- nested **groups** are inert: never displaced, never dropped, carried through

```
user.merge(wrapped)
  today:    throws filtersNotFlat
  proposed: and( eq(realmId,'ATTACKER'), and(eq(realmId,'SCOPE')) )   // intersection, scope survives

and(a, or(b,c)).merge(eq(b,9))
  today:    throws filtersNotFlat
  proposed: and( a, or(b,c), eq(b,9) )                                // no guessing inside the OR
```

Properties this buys:

- `merge` becomes **total**, so the throw disappears without giving anything up
- the guarantee no longer depends on callers not flattening, so `flatten()` is safe again
- **backward compatible** for the flat/flat case, which behaves identically
- **fail-closed**: a merge can only ever become narrower, never wider

Tradeoff: a condition inside a group can never be displaced, so overriding one means rebuilding rather than merging. That seems like the right default, since displacing inside a compound tree is precisely the operation with no sound semantics.

## Why now

Pre-GA. Both the error surface (`FILTERS_NOT_FLAT`) and the implicit "nesting means protected" contract become semver-frozen at 2.0.0.

## Consumer note

authup is not affected today. Nothing there merges filter trees: the entity collection manager strips filters before `mergeQueries` and combines them separately via `and()`, and the server injects scope with `appendQueryConditions` and passes the result straight to the adapter. The client-side filter is also not a security boundary, since the server re-injects scope independently. This is about the API contract rather than a live bug.

Verified against `@rapiq/core` 2.0.0-beta.15.

tada5hi:
Closes #875.

## The mechanism today

`and()` / `or()` produced a non-flat tree, and `merge()` refused anything that was not a flat root-AND. Tree **structure** carried two unrelated meanings at once: logical grouping, and "this condition is protected from displacement".

Two consequences, both described in #875:

1. `flatten()` collapses same-operator nests, so it turned a protected tree back into a displaceable one. The #871 guarantee held only as long as nobody normalized.
2. `merge()` was partial, and refused genuinely compound trees (`and(a, or(b,c))`, where per-field replace has no sound meaning) with the same error as wrap-injected ones (which are semantically a plain conjunction).

## What changes

Displaceability rides on the node as `ICondition.sealed`:

- `and()` / `or()` **seal** every condition they inject.
- A new `seal(condition)` helper exposes the marker on its own: immutable, idempotent, typed per node kind. It is what a filters `validate` hook reaches for when its policy residual has to survive composition.
- `flatten()` never hoists a sealed group, the same exception NOT already had.

`merge()` becomes **total**:

| Conjunct | Behavior |
|---|---|
| root-level leaf condition | replaced by a receiver condition on the same field (unchanged) |
| sealed condition | never replaced, carried through |
| nested group, non-AND root | never replaced, carried through as one conjunct |

```
user.merge(wrapped)          // and( eq(realmId,'ATTACKER'), eq(realmId,'SCOPE') <sealed> )
and(a, or(b,c)).merge(eq(b,9))  // and( a, or(b,c), eq(b,9) )
```

`MergeError.filtersNotFlat` and `ErrorCode.FILTERS_NOT_FLAT` are removed: composition narrows instead of failing, and a merge can no longer widen a result set.

Two shapes follow from putting the marker on the node instead of the nesting, both simplifications:

- an empty receiver no longer needs the double-nested group (`empty.and(scope)` is now `and(scope <sealed>)`),
- a receiver already carrying the injected operator contributes its conditions to the group instead of becoming a child, so the client's own conditions stay mergeable while the scope does not. The expression wire of a scoped query gets flatter as a result (`and(eq(level,'error'),eq(node_id,'node-1'))` instead of `and(and(eq(level,'error')),eq(node_id,'node-1'))`).

No `mergeOr` counterpart: an `or(...)` argument is carried as one inert conjunct and and-ed in, never turned into a root OR. Widening stays the explicit operation (`filters.or(...)`).

## Notes

- The seal is a **server-side composition marker, not wire grammar**. A sealed condition that is encoded and decoded again comes back displaceable, which is what the codec acceptance test now pins.
- The parser rebuild paths (`applyFiltersSchemaValidation`, relation pruning) preserve the marker when they reconstruct a node.
- Backward compatible for the flat/flat case, which behaves identically.

## Verification

`npm run test` (2011 tests across 10 projects), `npm run build`, eslint, docs build with anchor checks. Also replayed the exact reproduction from #875 against the built `dist` bundle: the scope survives both `merge()` and `flatten()` + `merge()`.

Docs updated: merging & composition (new `seal` section), errors, filters (`validate` hook), building queries, authorization recipe, `@rapiq/core` export table, plus `.agents/architecture.md`.

## Summary by CodeRabbit

- **New Features**
  - Added support for sealing filter conditions and groups so authorization and validation constraints remain protected during query merging and normalization.
  - Filter merging now preserves nested or protected conditions by combining them instead of rejecting them.
  - Added the public `seal()` helper for immutable, repeatable protection of filter conditions.

- **Documentation**
  - Updated query-building, filtering, authorization, and error guidance to explain protected conditions and revised merge behavior.
  - Clarified encoding behavior for protected compound conditions.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
Audit follow-up: filed #877 for a pre-existing interaction found while verifying this PR. `pruneFiltersByRelations` recurses into a sealed group, so a relations gate rejecting a relation also drops a policy residual that names it, leaving the query wider than the policy intended. It predates this branch (the residual was an unsealed group before), and the fix is a decision about parse-time pruning rather than about the merge law, so it stays out of here. Three options are written up in the issue.

tada5hi:
@coderabbitai ful lreview

tada5hi:
@coderabbitai pause

PR #879, opened 2026-08-03
fix(core): refuse to prune a sealed filter condition

### LINKED ISSUE #877: Relation pruning silently drops a sealed policy residual (fail-open)
Found while auditing #876 (sealed conditions). The behavior predates that PR: it is the interaction between the `relations` validate gate (#815) and a `filters` validate hook that returns a policy residual (#840). #876 does not introduce it, but it makes the contradiction explicit, since a sealed condition now carries "this must survive composition" on the node itself.

## What happens

A filters `validate` hook may answer with a condition that scopes the leaf it saw, the pattern documented in the [authorization recipe](https://rapiq.tada5hi.net/guide/recipes/authorization):

```js
filters: {
    allowed: ['name', 'realm.name'],
    validate: (filter) => (filter.field === 'name'
        ? seal(and(filter, eq('realm.id', 'SCOPE')))   // filter by name only within your realms
        : filter),
},
relations: {
    allowed: ['realm'],
    validate: (name) => name !== 'realm',             // this actor may not traverse realm
},
```

Client input: `{ filters: { name: 'John', 'realm.name': 'master' } }`

```
without the relations gate:
  and( sealed(and( name = John, realm.id = SCOPE )), realm.name = master )

with the relations gate rejecting `realm`:
  and( sealed(and( name = John )) )
```

The client's own `realm.name` leaf is dropped, which is exactly what the relations gate is for. But `pruneFiltersByRelations` recurses into the sealed group and drops the residual's `realm.id` leaf too, so the query comes back **wider than the policy intended**: it now returns Johns from every realm, silently.

## Precondition

Narrow. The relation has to be rejected, and the ledger only records relations that client input actually resolved, so all three must line up:

1. a filters `validate` hook whose residual references a relation path,
2. client input that names the same relation, and
3. a relations `validate` hook that rejects it.

A residual on a local column (`realm_id`) is never affected, which is why this has not surfaced in practice.

## Options

**A. Leave it, and document it.** If the actor may not traverse the relation, no condition on it can be enforced, so removing it is consistent with the relations gate. Cost: it fails open, silently, in the one place whose purpose is narrowing. The `seal` marker then means "survives composition, except pruning", which is the kind of qualifier #875 removed.

**B. Skip sealed conditions while pruning.** `pruneCondition` leaves sealed nodes intact. Cost: the surviving query names a relation the gate rejected, and adapters join what the query names, so the gate stops preventing the join. That trades a fail-open for a gate bypass, which seems worse.

**C. Throw a typed `ParseError` when pruning would drop a sealed condition.** New `ErrorCode` (e.g. `FILTERS_SEALED_PRUNED`). The two hooks contradict each other, which is a server misconfiguration rather than bad client input, and neither of the other outcomes is correct. Consistent with the fleet rule that a silent semantic downgrade loses to a loud failure. Blast radius is small given the precondition above.

Rejected while writing this: dropping the whole sealed group when any part of it is pruned. It removes the client leaf *and* the scope, which widens the result set further rather than narrowing it.

Recommendation: **C**, with the error naming both the pruned relation and the sealed condition, so the misconfiguration is obvious from the message.

## Reproduction

<details>
<summary>runnable script (built dist)</summary>

```js
import { SchemaRegistry, defineSchema, and, eq, seal } from '@rapiq/core';
import { SimpleParser } from '@rapiq/parser-simple';

function build(relationsOptions) {
    const registry = new SchemaRegistry();
    registry.add(defineSchema({ name: 'realm', filters: { allowed: ['id', 'name'] } }));
    registry.add(defineSchema({
        name: 'user',
        filters: {
            allowed: ['name', 'realm.name'],
            validate: (filter) => (filter.field === 'name' ?
                seal(and(filter, eq('realm.id', 'SCOPE'))) :
                filter),
        },
        relations: relationsOptions,
        schemaMapping: { realm: 'realm' },
    }));

    return registry;
}

const input = { filters: { name: 'John', 'realm.name': 'master' } };
const gated = build({ allowed: ['realm'], validate: (name) => name !== 'realm' });
const ungated = build({ allowed: ['realm'] });

console.log(JSON.stringify(new SimpleParser(ungated).parse(input, { schema: 'user' }).filters));
console.log(JSON.stringify(new SimpleParser(gated).parse(input, { schema: 'user' }).filters));
```
</details>

Relevant code: `pruneCondition` in `packages/core/src/parser/relation-prune.ts`.

tada5hi:
Closes #877 with **option C**: the contradiction between the two validators throws instead of failing open.

## The gap

A filters `validate` hook may answer with a sealed policy residual scoping the leaf it saw (the [authorization recipe](https://rapiq.tada5hi.net/guide/recipes/authorization) pattern). When that residual names a relation the relations `validate` hook rejects, `pruneCondition` recursed into the sealed group and dropped the residual along with the client's own leaf:

```
without the relations gate:
  and( sealed(and( name = John, realm.id = SCOPE )), realm.name = master )

before this PR, with the gate rejecting `realm`:
  and( sealed(and( name = John )) )        // Johns from every realm
```

## The fix

Pruning now treats a seal as protecting the whole subtree it heads. A drop inside (or of) a sealed condition throws, naming both sides of the contradiction:

```
SchemaError [schemaSealedConditionPruned]
The relations validator rejected "realm", but the sealed filter condition on "realm.id"
traverses it. A sealed condition must not be dropped, and a rejected relation must not be
joined: align the relations validator with the filters validator that sealed it.
```

An **unsealed** residual is an ordinary displaceable condition and is still pruned, which keeps `seal` the single marker that says "this must survive". Everything the gate did before is unchanged when no seal is in the way.

## `SchemaError`, not `ParseError`

The issue's option C said "typed `ParseError`", but also called this a server misconfiguration, and those point at different HTTP buckets in `errors.md` (`ParseError` maps to 400). Two contradicting hooks are a configuration bug on the receiving side, and a 400 would be swallowed as ordinary client noise. Precedent: `SCHEMA_VALIDATOR_ASYNC_REQUIRES_ASYNC_PARSER` is likewise detected during parse and classed as a schema error. New code: `ErrorCode.SCHEMA_SEALED_CONDITION_PRUNED`.

## Tests

- `packages/core/test/unit/parser/relation-prune.spec.ts`: sealed leaf, residual inside a sealed group, sealed condition below an unsealed group, sealed `elemMatch` (target and interior), and pruning that leaves an untouched seal alone.
- `packages/parser-simple/test/unit/parser/relations-traversal.spec.ts`: the issue's reproduction end to end (sync, async, standalone filters parse), plus the permitted-relation and unsealed-residual boundaries.

Docs updated: `errors.md` (code table), `relations.md` (rejection prunes deep), `filters.md` (validate replacements), `merging-queries.md` (seal section), `recipes/authorization.md` (warning box on scoping through a relation path vs. a local column).

## Summary by CodeRabbit

* **New Features**
  * Sealed filter conditions are now protected during relation validation.
  * Parsing reports a clear schema error when authorization would remove a required sealed condition, including the affected relation and field.
  * Unsealed conditions continue to be pruned normally, while sealed defaults receive the same protection.

* **Documentation**
  * Clarified sealed-condition behavior, error handling, and recommended configuration.

* **Tests**
  * Added coverage for nested, element-match, synchronous, asynchronous, standalone, and default-filter scenarios.


tada5hi:
## Self-audit follow-up (eef50407)

Probed the edges rather than trusting the happy path. Three findings, all addressed.

### 1. Sealed schema `default` was input-dependent (real, introduced by the first commit)

`filters.default` is materialized in two different places: in the parser's `build()` when the client sent no filters (so it goes *through* pruning), and in `pruneFiltersByRelations`' fallback when pruning emptied the tree (so it does not). With a sealed default naming a rejected relation, that meant:

```
sealed default | realm named via relations  => THROWS
sealed default | realm named via own filter => survives silently
```

The same schema and the same actor, decided by what the client happened to send. The fallback now runs the same check, so a sealed default behaves like a sealed residual in every shape. **Unsealed defaults keep surviving unchanged**: they are the server-authored baseline, and dropping them would be the fail-open this PR is closing.

### 2. The "pruning would widen" justification was only true for the `and` shape

Dropping an arm of a sealed `or` *narrows*, and the docs/jsdoc claimed otherwise. The refusal is still correct there (the sealed condition is not honored as written, and a policy grant losing an arm denies legitimate access), but it is a per-node rule, not an operator-aware one. Wording corrected, and the `or` / `not` shapes are now pinned by tests so the stricter-than-fail-open behavior is deliberate rather than incidental.

### 3. Verified, no change needed

- **Dialect parity**: expression and mongo parsers throw identically (they share the core helper).
- **No swallowing**: no `catch` on the parse or schema-aware encode path, so the error propagates instead of degrading to a silent drop.
- **Post-parse composition unaffected**: `query.filters.and(scope)` runs after pruning, so server scoping never trips this.
- **Vacuous nodes** (`seal(and())`, a sealed `elemMatch` with an empty interior) are still dropped silently. They carry no restriction and no parser produces them.
- **Fields**: a gated `Field.condition` is unaffected, since pruning drops the whole field and the column is never projected. No fail-open there.

### Known cost, unchanged from the issue's analysis

Once the two hooks contradict each other, the failure is **client-triggerable**: any request naming the rejected relation now 500s instead of silently widening. That is the accepted trade of option C (the misconfiguration has to be visible to the operator), and it is bounded by the same three preconditions the issue lists.

tada5hi:
## Recipe change: seal the residual, not the group (7b59c773)

Follow-up to the audit. The blast radius of the new error was wider than #877 assumed, and the cause was the recommended shape rather than the rule.

The issue notes that "a residual on a local column (`realm_id`) is never affected". With `seal(and(filter, <scope>))` that is not true, because a seal protects the whole subtree it heads, including the client's own leaf:

```
residual on realm_id, client sends filter[realm.name]=master, relations hook rejects realm

seal(and(leaf, residual))  => THROWS
and(leaf, seal(residual))  => [ realm_id = S (sealed) ]     leaf dropped, scope kept
```

The first case is not a misconfiguration: the residual is fine, the gate is fine, and they collide only because the seal happens to cover the client's leaf. The second is what you want, and it gives up nothing:

```
flatten()          : [ name=John, realm_id=S (sealed) ]     marker survives hoisting
merge, either side : group carried through, realm_id=ATTACK and-ed in alongside
```

A group is never displaced as a unit either, so both shapes resist a merge; the seal only matters once `flatten()` hoists the residual into the root, and that works on the leaf.

So the recipe now reads `and(filter, seal(inArray('realm_id', actor.realmIds)))`, updated in the filters guide, the authorization recipe, the seal section of the composition guide and `.agents/architecture.md`. Three tests pin the recommended shape: the client leaf prunes away cleanly, the residual keeps its marker through `flatten()`, and a residual that *itself* names a rejected relation still throws.

Net effect: the error now fires only on the contradiction #877 actually describes.

PR #890, opened 2026-08-07
fix(core)!: make query filter merges conjunctive

### LINKED ISSUE #889: Pre-GA: is displaceability worth its contract cost, or should IR filter composition just be conjunctive?
Raising this before v2 goes GA, because it is a contract question rather than a defect: the displaceability machinery may be defending a primitive that consumers do not want, and beta.18 moved its cost onto the brand new extensibility surface.

Not a bug report. Everything below behaves as documented.

### What seal currently buys

The chain for a server-injected scope is:

`IFilters.and` seals the injected condition -> the seal survives `flatten` -> it therefore stays a nested group -> a later `merge` leaves it alone.

Worth noting the last hop is not the seal's doing. `IFilters.merge` already documents that **a sealed condition and a nested group are both inert**. So for the wrap-and-inject path the group structure alone would survive a merge; the seal's unique contribution is stopping `flatten` from hoisting that group in the first place.

### Why this looks like it is defending an unwanted primitive

Displaceability only exists because `IFilters.merge` does per-field replace, where a conjunct can be dropped because the other side constrains the same field. That is a reasonable thing to want when a caller overrides a default while *building* a query. It is a hazardous thing to keep available on the IR, which is where server-derived authorization scopes live.

The evidence from the consumer side is that the primitive is being routed around rather than used. In authup, the entity collection manager calls `stripFilters()` on **every** query before handing it to `mergeQueries`, and composes filters separately through `and()`:

```ts
const chrome = mergeQueries(stripFilters(propsQuery), stripFilters(contextQuery));

return new Query({
    fields: chrome.fields,
    relations: chrome.relations,
    sorts: chrome.sorts,
    pagination: chrome.pagination,
    filters: combineScopedFilters(propsQuery.filters, contextQuery.filters),
});
```

So filters never enter a merge anywhere in the codebase. Fields, relations, sorts and pagination merge happily; filters were carved out because per-field replace on an authorization-bearing tree was not the wanted semantic. The carve-out is the finding: the one place the library offers filter composition, the consumer declined it and hand-rolled conjunction.

Two things follow. The seal is defending a path that is not taken, and `mergeQueries` is less useful than it looks, since a consumer holding server-derived filters has to decompose it and reassemble the result by hand.

### What it costs

beta.18 made `seal()` a required member of `ICondition`, so every custom condition must now implement immutable sealing. That tax lands on the extensibility surface introduced in the same release: an author adding a `jsonContains` operator has to reason about displaceability before their condition composes, in order to defend against a merge they may never perform.

The marker is also same-process only, as the docs are explicit about:

> The seal is a server-side composition marker, not part of any wire grammar: a sealed condition that is encoded and decoded again comes back displaceable.

So across an encode/decode boundary the protection is carried entirely by the nesting, which merge treats as inert anyway.

### The alternative

Make IR filter composition monotonic. Either drop filters from `mergeQueries` (matching what consumers already do by hand), or make `IFilters.merge` conjunctive so nothing can displace anything.

Per-field replace would move to the build-input layer, before a `Query` exists, which is where override-a-default actually belongs. On the IR, composition would only ever narrow.

That removes the displaceability axis outright: no marker, no `sealed` flag on the contract, no flatten-hoisting hazard, no `FILTERS_NOT_FLAT` (already gone), no `SCHEMA_SEALED_CONDITION_PRUNED`. `filters.merge()` appears to have exactly one internal call site (`mergeQueries`), so the blast radius looks small.

### Arguments against, which I do not think are decisive but are real

**Pruning.** `pruneFiltersByRelations` raising `SCHEMA_SEALED_CONDITION_PRUNED` on a sealed subtree is a genuine correctness win, independent of merge: it surfaces a contradiction between the relations gate and the filters validator that would otherwise fail open or closed depending on tree shape. If the marker stays, this is its strongest justification, not merge. It may also be expressible over nested groups without a per-condition flag.

**The Fields asymmetry is deliberate.** `Fields.merge` throws `MergeError` rather than sealing, and the code explains why: a filter can be and-ed in while a field is either gated or not. Fair. But that reasoning argues for filters composing conjunctively, which is the proposal here, rather than for a displaceability marker.

**Someone may want replace semantics on filters.** Likely true for query building. The claim is only that it belongs before the IR, not on it.

### Suggested shape, if the argument lands

1. Make IR filter composition conjunctive, or remove filters from `mergeQueries`.
2. Drop `sealed` / `seal()` from the public `ICondition` contract so custom conditions do not pay for it. Keep it internally if pruning still wants it.
3. Keep per-field replace at the build-input layer.

Happy to prototype (1) against authup, which exercises server-injected realm and ownership scopes through `appendQueryConditions` on every list endpoint and would notice a regression quickly. Flagging it now because dropping a member from `ICondition` is a breaking change, and the window for that closes at 2.0.0.

tada5hi:
## Summary

- make `mergeQueries` compose filters as an ordered logical `AND` so every predicate survives, including same-field ranges and contradictions
- separate live condition identity from relation-pruning protection with `CONDITION_MARKER` and `preserve()`
- remove the public `seal`, `sealed`, and `seal()` contracts
- keep preserved policy residuals fail-closed through validation, `elemMatch`, relation pruning, and schema defaults
- update public and agent documentation for conjunctive composition and pre-IR replacement

## Breaking changes

- same-field filters no longer replace one another during `Filters.merge()` or `mergeQueries()`; they remain conjuncts
- custom conditions must extend `Condition` or provide `CONDITION_MARKER`
- `seal()`, `sealed`, and condition seal methods are removed
- `SCHEMA_SEALED_CONDITION_PRUNED` is renamed to `SCHEMA_PRESERVED_CONDITION_PRUNED`
- the deprecated simple URL codec can reject a merged tree with duplicate fields or compound filters; the expression codec supports these trees

## Validation

- `npm run lint -- --ignore-pattern '.claude/worktrees/**'`
- `NX_SKIP_NX_CACHE=true npm run build`
- `NX_SKIP_NX_CACHE=true npm test` (2,132 tests)
- `git diff --check origin/master...HEAD`

Closes #889

## Summary by CodeRabbit

* **New Features**
  * Added `preserve()` for protecting filter conditions during validation and relation pruning.
  * Added `mergeFiltersInput()` for predictable, field-level filter input merging.
  * Improved recognition of custom query conditions.
* **Behavior Changes**
  * Filters now combine using ordered logical AND, retaining all predicates.
  * Server-provided conditions remain conjunctive with client filters.
  * Other query parameters retain left-priority merging.
* **Breaking Changes**
  * Replaced `seal()` and sealed-condition terminology with preservation semantics.
  * Renamed the related schema error and public error method.
* **Documentation**
  * Updated guides, examples, and migration guidance.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #895, opened 2026-08-10
feat: schema index declarations (indexed filters and sort)

tada5hi:
## Summary

Adds a schema-level `indexes` declaration that restricts which filter combinations and sort key sequences a client may request, enforced at parse time. Replaces sort tuple groups (`allowed: string[][]`) outright, which is free pre-GA.

Design spec: `.agents/plans/029-schema-indexes.md` (local).

## API

```typescript
defineSchema<User>({
    indexes: [
        ['realm_id', 'created_at'],
        ['email'],
    ],
    filters: { indexed: true },     // true | 'anchor' | 'cover'
    sort: { indexed: true },
});
```

## Semantics

- **Filters, anchor mode** (default): every AND group of the final parsed tree must contain at least one conjunct whose field leads a declared index; the rest is residual filtering. An `or(...)` passes only when every branch passes (index-merge reality); a compound conjunct anchors by passing itself.
- **Filters, cover mode**: additionally, per relation path, the AND group's field set must equal a leftmost prefix (as a set) of one index.
- **Sort**: requested keys must equal, in order, a leftmost prefix of one index; directions ignored; keys must share one relation path (no single index serves cross-table ordering).
- **Structural only**: operators, negation and case folding are deliberately ignored; sargability differs per engine (trigram GIN serves `contains`, hash serves only equality), so rapiq trusts the declaration instead of embedding a B-tree capability table.
- **Dotted keys** check the target schema's own declaration (registry + `schemaMapping`).
- The check runs on the **final tree** (after `validate` hooks and relation pruning): server-authored policy residuals legitimately anchor the executed query; `default`s bypass.
- **Failure** follows the standard drop/throw policy with new `ErrorCode.KEY_COMBINATION_NOT_INDEXED`; drop is whole-parameter (an index violation is a combination property, dropping single leaves could drop the anchor). Two hardening rules on the filters drop path (added after an adversarial review pass): a violating tree carrying a preserved condition refuses to drop it and throws `SCHEMA_PRESERVED_CONDITION_PRUNED`, and a violation on a schema without a filters `default` always throws, since dropping to an empty filter set would execute exactly the unfiltered scan the policy prevents. The defaults bypass compares by semantic value, so codec round trips of a defaults tree survive.

## Placement

Pure check functions in core (`schema/indexes/`), policy appliers (`parser/index-policy.ts`) invoked from `BaseQueryParser` (composed path, all dialects) and each standalone `parse()` tail (parser-simple filters+sort, parser-expression, parser-mongo). Adapters, build layer, merge and codec-url need no changes; codec-url enforcement composes through the schema-bound decoder (pinned by a new spec).

## Breaking (pre-GA)

- `sort.allowed` narrows to a flat `string[]`; tuple groups removed (parser block, `ResolutionScope` special case, types). Migration note added to the v1 migration page: declare `indexes` + `sort: { indexed: true }` instead.
- `describe()` gains `indexes` (top-level), `filters.indexed`, `sort.indexed`.

## Tests

- core: 37-case spec for the two pure check functions (anchor/cover, OR/NOT, nested flatten, preserve transparency, custom-condition skip, relation paths); policy applier spec (drop/throw/defaults bypass).
- parser-simple/-expression/-mongo: standalone + composed enforcement specs (drop, typed throw, validate-residual anchoring, relation anchor, sort prefix).
- codec-url: decode enforcement spec, no src changes.
- Full monorepo build and test green (11 projects / 10 suites).

## Summary by CodeRabbit

* **New Features**
  * Added schema-declared indexes for validating filter combinations and sort keys.
  * Added anchor and cover validation modes, including relation-aware index checks.
  * Invalid filters and sorts now fall back to schema defaults or raise a clear parse error when configured.
  * Added indexed parsing support across expression, Mongo, and simple parsers.

* **Documentation**
  * Documented index configuration, indexed filtering and sorting, errors, and migration guidance.

* **Breaking Changes**
  * Sort allow-lists now use flat key lists; composite ordering is configured through schema indexes.


tada5hi:
@coderabbitai pause

PR #902, opened 2026-08-12
feat(adapter-typeorm): assert schema indexes against entity metadata

### LINKED ISSUE #898: adapter-typeorm: assertSchemaMatchesEntity should cover schema `indexes` declarations
## Use case (authup)

With 2.0.0-beta.20's schema index declarations (#895), authup declares `indexes` on all 26 registered entity schemas and opts into `filters: { indexed: true }` + `sort: { indexed: true }`. The declarations are structural promises — rapiq trusts them — so drift between a declared sequence and the actual database index set silently turns the index policy into a lie (a promised combination stops being index-served, or an honest one starts throwing after an entity refactor).

`assertSchemaMatchesEntity` already validates allow-lists, fields/sort defaults and the filters default tree against TypeORM `EntityMetadata`, but does not look at `indexes`.

## Proposal

Extend `assertSchemaMatchesEntity` so every declared index sequence must be a **leftmost prefix** of one of:

- `metadata.primaryColumns` (property-name order),
- a unique constraint (`metadata.uniques[].columns`),
- an index (`metadata.indices[].columns`),

comparing property-name sequences (schema `indexes` hold resolved/mapped names, which for a mapping-less schema are the property paths). Prefix rather than equality: a declared `[a]` is served by a real `(a, b)` index.

## Interim

authup carries a local `assertSchemaIndexesMatchEntity` with exactly that rule in its boot-time validation pass (`apps/server-core/src/app/modules/database/repositories/schema-validation.ts`); it should be deletable once this lands.

tada5hi:
Closes #898.

`assertSchemaMatchesEntity` validated allow-lists, fields/sort defaults and the filters default tree, but ignored the `indexes` declaration added in #895. Since rapiq trusts that declaration structurally, drift between it and the real database index set silently turns an `indexed` filters/sort policy into a lie.

## The rule

Every declared sequence must be a **leftmost prefix** of one of:

- `metadata.primaryColumns` (property-name order),
- a unique constraint (`metadata.uniques[].columns`),
- an index (`metadata.indices[].columns`).

Prefix rather than equality: a real `(realm_id, email)` index also serves a declared `['realm_id']`. Both metadata lists are consulted because the driver decides where a constraint lands: the mysql family stores `@Unique` as a unique index, cockroachdb stores a unique `@Index` as a unique constraint (pinned by a spec that builds mysql metadata).

## Failure reporting

Index keys enter the existing column-key check first, so a renamed column keeps surfacing as `SchemaEntityMismatchError`. A sequence of live columns that no real index leads throws the new `SchemaEntityIndexMismatchError` (`ErrorCode.SCHEMA_ENTITY_INDEX_MISMATCH`, a `SchemaError` subclass) carrying `schema`, `entity` and **every** offending sequence:

```
The schema "user" declares indexes not backed by the entity User: (last_name), (email, realm_id).
```

A schema without an `indexes` declaration (or with `indexes: []`) skips the check.

## Notes

- `errors/module.ts` is split into one file per error class (`schema-entity-mismatch.ts`, `schema-entity-index-mismatch.ts`), mirroring core's `errors/` layout; the lint config caps a file at one class.
- The `User` test entity gains `@Index(['realm_id', 'email'])` so the composite-index path is covered next to the existing `@Unique(['first_name', 'last_name'])`.
- Docs: a "Declared indexes" section in the adapter-typeorm page, the new code in the error-handling table and hierarchy, and a pointer from the schema `indexes` guide. The stale "(tuple groups included)" mention in the validated-keys list is dropped.
- authup's interim `assertSchemaIndexesMatchEntity` can be deleted once this ships.

## Summary by CodeRabbit

* **New Features**
  * Added validation for declared database indexes against entity definitions.
  * Reports detailed errors for unsupported, invalid, or unbacked index definitions.
  * Added a dedicated error code for index mismatches.

* **Documentation**
  * Expanded schema and error documentation with index validation rules and troubleshooting details.

* **Tests**
  * Added coverage for valid indexes, invalid ordering, missing columns, relation fields, and database-specific metadata.


tada5hi:
@coderabbitai review

PR #906, opened 2026-08-12
feat: accept sorts as the canonical sort input key, reject unknown build and schema keys

### LINKED ISSUE #905: core: defineQuery silently ignores unknown top-level build-input keys, yielding an unfiltered query
## What happens

`defineQuery` drops any top-level key it does not recognise, without a warning or error. The failure mode is the dangerous direction: a mistyped filter key produces an **empty query**, which downstream means *unfiltered data*, not *no data*.

Measured through `@rapiq/codec-url`'s encoder (rapiq 2.0.0):

```ts
buildQueryString({ filters: { active: false } })   // "?codec=url-expression&filter=eq(active,'false')"
buildQueryString({ filter:  { active: false } })   // ""   <- silently unfiltered
buildQueryString({ filter:  { name: 'foo' } })     // ""   <- same, nothing to do with the value
```

The `filter` / `filters` confusion is an easy one to hit because it is the **wire** key: the URL carries `?filter[name]=x` (singular) while the build input takes `filters` (plural). A v1 habit, a hand-written JS caller, or a TypeScript cast all land here.

## Why it matters downstream

In authup a query object flows into an authorization-adjacent read path. A dropped filter is not a cosmetic bug there: `getMany({ filter: { userId } })` intending "this user's rows" silently becomes "the first page of everyone's rows". Server-side allow-lists and per-row gates still apply, so it is not a privilege escalation — but it is a correctness and data-exposure-shaped footgun, and it is invisible in review because the call looks right.

I hit it writing a test, with the mistake masked by an `as any`. TypeScript does reject it when the input is properly typed, so this is specifically about runtime robustness for JS consumers, casts, and inputs crossing a boundary.

## Suggestion

Reject (or at minimum warn on) unknown top-level keys in `defineQuery` / the build layer — the same posture the **decode** side already takes after #883 and #893, where a hostile or unexpected key raises a typed `ParseError` rather than being ignored. A `strict` option would be enough if a hard throw is too breaking, though given the failure mode I would argue for throwing by default pre-GA... which this now is not, so an option plus a documented note may be the pragmatic call.

Related: tada5hi/rapiq#898, tada5hi/rapiq#904 (same downstream adoption).

tada5hi:
`sort` was the only query parameter whose AST property (`Query.sorts`) differed from its input key (`sort`), while the build type was already called `SortsBuildInput` and the schema classes were singular (`SortSchema`) where every sibling is plural (`FieldsSchema`, `FiltersSchema`, `RelationsSchema`).

This makes `sorts` the canonical spelling on every developer-authored input surface, keeps `sort` working as `@deprecated` (removal in 3.0), and folds in #905.

**The URL wire parameter does not change: the query string still carries `sort=-id`**, exactly as the `filters` input key is carried as `filter`.

## What changed

| Surface | Before | Now |
|---|---|---|
| `defineQuery` build input | `sort` | `sorts` canonical, `sort` deprecated |
| `defineSchema` options | `sort` | `sorts` canonical, `sort` deprecated |
| `Schema` property | `.sort` | `.sorts` canonical, `.sort` the identical instance |
| `parse()` input | `sort` | either spelling |
| `ParseQueryOptions` skip flag | `sort` | either spelling |
| typeorm `EntitySchemaOptions` | `sort` | `sorts` canonical, `sort` deprecated |
| prisma `ModelSchemaOptions` | `sort` | `sorts` canonical, `sort` deprecated |
| `describe()` output | `sort` | both `sorts` and `sort` |
| URL wire | `sort` | `sort` (unchanged) |

`Parameter` gains `SORTS = 'sorts'`; `Parameter.SORT` keeps its **value** `'sort'` and is deprecated, so `parameters: ['sort']` masks, `describe()` readers and error payloads all keep working. Both spellings are accepted in every parameter mask.

The sort schema exports are now plural (`SortsSchema`, `defineSortsSchema`, `SortsOptions`, `SortsOptionDefault`, `SortsSchemaDescription`), with the singular names kept as deprecated aliases of the same values.

Supplying **both** spellings throws a typed `KEY_AMBIGUOUS` error rather than merging or silently picking a winner. The throw is unconditional, not gated by `throwOnFailure`, and applies on all three surfaces. An `undefined` side never triggers it, so a spread migration wrapper (`{ sorts: props.sorts, sort: props.sort }`) is safe.

## Closes #905

`defineQuery` dropped unknown top-level keys silently, so `defineQuery({ filter: ... })` (the wire spelling) produced an empty and therefore **unfiltered** query. Unknown top-level keys now raise a typed error with a did-you-mean pointer:

```
defineQuery({ filter: { active: false } })
// BuildError: The key filter is unknown. Did you mean filters?
```

The same guard covers `defineSchema`, where a mistyped key (`field:` instead of `fields:`) silently declared no allow-list at all, the more permissive direction. Suggestions cover the wire names and obvious singulars: `filter` to `filters`, `page`/`limit`/`offset` to `pagination`, `include`/`relation` to `relations`, `field` to `fields`.

`parse()` input deliberately still ignores unrelated keys: a decoded URL legitimately carries tokens and custom parameters.

## Release notes

This must ship as a **minor, never a patch**. The two `fix(core):` commits make previously-working calls throw: `defineQuery(someObject)` carrying unrelated keys used to drop them. Do not cherry-pick `8ac9bcc7` or `1f8e337f` onto a patch line.

Three observable changes worth calling out:

1. **Unknown build/schema keys now throw.** Loud, immediate, first-call, with a suggestion. TypeScript does not catch the real-world case, since excess-property checking only fires on fresh literals and #905 is about a variable being passed, so the runtime guard is the only line of defence. Strictly better than the silent-unfiltered-data outcome it replaces, but it lands in the changelog under "Bug Fixes" and deserves a note.
2. **`scope.parameter` is now `'sorts'`** for a sort `validate`/`validateMany` hook, where it was `'sort'`. Anything branching on that value needs updating.
3. **`describe()` output carries a new `sorts` key.** Any downstream deep-equality snapshot of a schema description needs updating.

Apps calling a parser directly on client data now have a client-triggerable `ParseError` via a body carrying both `sort` and `sorts`. The documented recipe already maps `ParseError` to a 400, so it degrades correctly.

## New public API

`Parameter.SORTS`, `ErrorCode.KEY_UNKNOWN`, `ErrorCode.KEY_AMBIGUOUS`, `normalizeParameter`, `resolveAliasedKey`, the five plural sort schema exports, and the `sorts` keys on `QueryBuildInput`, `SchemaOptions`, `ParseQueryOptions`, `SchemaDescription`, `EntitySchemaOptions` and `ModelSchemaOptions`. `assertKnownInputKeys` and `suggestInputKey` stay internal.

## Verification

Sixteen commits, each with its own tests, reviewed per commit plus a whole-branch review. `nx run-many -t build` 11/11, `npm run test` 10/10 projects (2302 tests), `npm run lint` clean, docs site builds.

Invariants checked empirically rather than from prose:

- encoding through both dialects, plain and schema-aware, still emits `sort=-id` and never `sorts=`
- `Parameter.SORT === 'sort'` unchanged
- every `sort` spelling accepted at 2.0.0 is still accepted
- `schema.sorts === schema.sort` (the same instance, since `extendSchemaOptions`/`setIndexes` mutate it in place)
- `{ sorts: false }` and `{ sort: false }` both skip; `false` is not swallowed by the definedness check
- `resolveAliasedKey` reads own properties only, so an array input does not see `Array.prototype.sort`

## Follow-up

A 3.0 issue should track removing the `sort` input key on every surface, `Parameter.SORT`, the `describe()` `sort` key and the singular class aliases. The still-singular internal vocabulary belongs in that same pass: `IRootAdapter.sort`, `parser.parseSort()`, `decodeSort()`/`encodeSort()`, `SortParseError`, `SortParseOptions`.

## Summary by CodeRabbit

- **New Features**
  - Introduced the canonical `sorts` query and schema property across the platform.
  - Added support for the deprecated `sort` alias for backward compatibility.
  - Added clear errors for unknown keys and conflicting `sorts`/`sort` usage.
  - URL encoding continues to use the compatible `sort` wire parameter.

- **Documentation**
  - Updated guides, examples, and adapter documentation to consistently describe `sorts`.
  - Documented migration guidance, alias behavior, and new validation errors.


tada5hi:
@coderabbitai pause

PR #907, opened 2026-08-12
feat(core): enumerate registered schemas via SchemaRegistry.getAll()

### LINKED ISSUE #904: core: SchemaRegistry exposes no way to enumerate registered schemas
## Use case (authup)

With schema `indexes` + `indexed` enforcement adopted, authup wants a test that walks **every registered schema** and asserts the invariant its declarations rest on: every allowed filter key and every allowed sort key must LEAD a declared index. A key that leads nothing is a hard 400 the moment a client filters or sorts by it alone (no filters default -> the drop path escalates to a throw), so this is worth pinning rather than reviewing by hand.

`SchemaRegistry` keeps `entities` `protected` and exposes only `get`/`getOrFail`/`resolve`/`add`/`drop` — there is no iterator, no `getAll()`, no `names()`. So a consumer that has handed its schemas to a registry cannot get them back out as a set, and has to keep (and export) a parallel array purely for iteration:

```ts
// core/query/module.ts — exported ONLY so a test can enumerate
export const schemas: Schema<any>[] = [clientSchema, roleSchema, /* ...24 more */];
for (const schema of schemas) schemaRegistry.add(schema);
```

## Proposal

Any one of these would do:

- `getAll(): Schema[]` (or `values()`),
- `names(): string[]`,
- or making `SchemaRegistry` iterable (`[Symbol.iterator]`).

Read-only enumeration seems uncontroversial — the map is already the registry's whole purpose, and `get()` hands out the same objects one at a time.

## Related

Same downstream adoption as tada5hi/rapiq#898 (index validation in `assertSchemaMatchesEntity`). If that lands, the boot-time pass would want enumeration too, rather than authup's hand-maintained schema-name -> entity-class map.

tada5hi:
Closes #904.

`SchemaRegistry` kept `entities` protected and exposed only `add`/`drop`/`get`/`getOrFail`/`resolve`, so a consumer that handed its schemas to a registry could not get them back out as a set. authup had to keep and export a parallel array of 26 schemas purely so a test could iterate them.

```typescript
getAll() : Schema<any, CONTEXT>[] {
    return Array.from(this.entities.values());
}
```

## Semantics

- **Registration order** (Map insertion order). Re-registering a name replaces the entry in place and keeps its position; a `drop()` followed by an `add()` moves it to the end. Both are a documented guarantee pinned by tests, since a consumer deriving `it.each` titles from the order relies on a contract rather than an accident.
- **Snapshot array, live elements.** The array is a fresh copy the caller owns: sorting or splicing it changes nothing in the registry, and a later `add()`/`drop()` leaves an array already held untouched. There is no concurrent-modification hazard by construction. The elements are the same instances `get()` returns, which is what the real consumers need: both feed the enumerated value straight back into `decodeQuery(..., { schema })` and into `assertSchemaMatchesEntity(schema, metadata)`, so identity matters and names or descriptions alone would not do.
- **Never throws.** An empty registry yields `[]`. No new `ErrorCode`, no error-class factory, `errors.md` untouched.
- **Encapsulation unchanged.** `entities` stays `protected` and the Map is never handed out; there is no write path through the return value. Nothing in the monorepo subclasses `SchemaRegistry`.

## Why this shape

- **Values only.** Names are derivable from the elements. A `[name, schema]` pair form would charge destructuring at every name-free call site to buy away one `as string` at the name-keyed ones, and it would write the key-vs-`schema.name` divergence (reachable only through a post-add rename that no consumer performs) into a GA contract. That divergence is a `Schema` typing wart and belongs upstream, not in the registry's collection API.
- **Not `values()`/`keys()`/`entries()`/`size`/`has()`/`forEach()`.** The mutators are deliberately `add`/`drop` rather than `set`/`delete`; adopting Map vocabulary here invites the rest of the Map shape. `.length` and `get()` already answer `size` and `has`, and an owned array already has `forEach`.
- **Not `[Symbol.iterator]` alone.** #904 is a discovery failure, not a capability gap: the parallel array exists because nothing on the class looked enumerable, and a symbol-keyed member appears in no `registry.` member completion. The choice stays monotonic either way: `[Symbol.iterator]() { return this.getAll()[Symbol.iterator](); }` remains a one-line additive follow-up if iterable interop is ever wanted.
- **`getAll` as the name.** `getOrFail` already establishes get-plus-qualifier as a second read path over the same map, so `getAll` is that construction applied to the plural case: bare verb, house accessor style, and it sorts beside `get`/`getOrFail` in completion.
- **Return type is `Schema<any, CONTEXT>[]`, not generic.** `get`/`getOrFail` take a `<T extends ObjectLiteral>` because the caller names one schema and knows its record type; a registry holds a heterogeneous set, so `getAll<User>()` would misdescribe every other entry. `CONTEXT` is threaded through, as `get`/`getOrFail` do (`resolve()`'s CONTEXT drop is a pre-existing inconsistency and is not copied). For the record, the tempting justification that a `<T extends ObjectLiteral = ObjectLiteral>` mirror would force casts on consumers was measured false; the objection is honesty about the set, not assignability.

## Changes

- `packages/core/src/schema/registry/module.ts`: the method plus its TSDoc. No new exported type, no barrel edit.
- `packages/core/test/unit/schema/registry.spec.ts`: a `getAll` block covering empty registry, registration order, identity with `get()`, re-add keeping position, drop-then-add appending, snapshot isolation from later mutation, and that mutating the returned array does not write back.
- `packages/docs/guide/schemas.md`, `packages/docs/packages/core.md`: the registry method list plus the semantics and an index-sweep example.
- `.agents/architecture.md`: one line recording the values-only decision so it is not re-litigated.

## Verification

`tsc --noEmit` clean, `getAll` present in the emitted `index.d.mts`, eslint clean on the changed files, all ten packages' test targets pass, docs site builds.

## Notes

Purely additive; release-please's linked group bumps all ten packages to 2.1.0.

Two follow-ups deliberately left out of this PR:

1. `Schema.name` is typed `string | undefined` even though `add()` refuses a nameless schema, so name-keyed callers write `as string`. That is a `Schema` typing decision, semver-relevant across the linked group, and must not be solved through the registry's collection API.
2. `defineSchemaRegistryWithDataSource` (adapter-typeorm) and `defineSchemaRegistryWithDatamodel` (adapter-prisma) each keep a local `names` Set and probe with `registry.get(name)`. `getAll()` is the primitive an `assertSchemaRegistryMatchesDataSource`-style boot-time helper needs; it can now ship with no further core surface.

## Summary by CodeRabbit

* **New Features**
  * Added `SchemaRegistry.getAll()` to retrieve all registered schemas in registration order.
  * Results are returned as a safe snapshot while preserving the original schema instances.
  * Supports inspecting schema details such as descriptions and indexes.

* **Documentation**
  * Added usage guidance and examples for retrieving registered schemas.


tada5hi:
@coderabbitai pause

PR #908, opened 2026-08-12
feat(core): guarantee a name at registration, narrow the enumerated element

tada5hi:
Follow-up to #907, now that it has merged: this targets `master` with the latter merged in.

`SchemaRegistry.add()` already refused a schema without a name at runtime, but nothing carried that guarantee outward: `Schema.name` is typed `string | undefined`, so a consumer keying a lookup off an enumerated schema had to write `ENTITY_TARGETS[schema.name as string]` for a case registration had already ruled out. This closes that, and fixes a latent bug found on the way.

## 1. `fix(core)`: renaming a schema left its sub-schemas behind

`Schema` stamps its name onto the five sub-schemas in `extendSchemasOptions()`, which runs once, from the constructor. The `name` setter inherited from `BaseSchema` is public, so assigning it afterwards desynced them:

```typescript
const schema = defineSchema({ name: 'before' });
schema.name = 'after';

schema.name;                    // 'after'
schema.fields.name;             // 'before'   ← stale
```

That is not cosmetic. A sub-schema reaches back into the registry by the name it carries: `ResolutionScope.for()` does `registry.get(parameterSchema.name)` (`resolver/module.ts:251`) and `resolveBase()` does the same (`:637`). A stale name resolves to the schema this one used to be, or to nothing.

The setter is overridden on `Schema` to restamp the sub-schemas, including when the name is unset.

## 2. `feat(core)`: `add(name, schema)`

```typescript
registry.add(userSchema);           // uses the declared name, throws without one
registry.add('item', itemSchema);   // schema adopts the name, wins over a declared one
```

The schema adopts the passed name (and, via part 1, restamps its sub-schemas), so the registry key and the schema's own name cannot disagree. The single-argument form is unchanged and still throws `SchemaError.nameUndefined()`. Both paths end at the same invariant: everything in a registry carries a `string` name.

## 3. `feat(core)`: `RegisteredSchema`, and `getAll()` narrowed to it

```typescript
export type RegisteredSchema<RECORD extends ObjectLiteral = ObjectLiteral, CONTEXT = any> =
    Schema<RECORD, CONTEXT> & { name: string };

getAll() : RegisteredSchema<any, CONTEXT>[]
```

So the enumerate-then-correlate consumer this whole line of work is for reads:

```typescript
for (const schema of registry.getAll()) {
    const entity = ENTITY_TARGETS[schema.name];   // string, no assertion
}
```

Deliberate limits:

- **One cast, inside `getAll()`.** The compiler cannot see the runtime guard in `add()`, so the guarantee is asserted in exactly one place rather than checked. Stated in a comment there.
- **`get()`/`getOrFail()` are not narrowed.** Both pass a `Schema` instance argument straight through without registering it, so their result is not necessarily a registered schema.
- **The guarantee covers registration, not later mutation.** `name` stays settable, and setting it does not re-key the entry, so `get()` keeps resolving under the name the schema was registered with. Documented next to the accessor.
- One ergonomic cost worth naming: a `RegisteredSchema[]` will not accept a bare `Schema` pushed into it. Correct, but it does surface at call sites that build lists from mixed sources.

## Changes

- `packages/core/src/schema/module.ts`: overridden `name` accessor pair + `propagateName()`.
- `packages/core/src/schema/registry/module.ts`: `add()` overloads, `getAll()` return type.
- `packages/core/src/schema/registry/types.ts` (new) + barrel line: `RegisteredSchema`.
- `packages/core/test/unit/schema/schema.spec.ts`: a `name` block covering rename, first assignment and unset propagation (all three fail on the parent commit).
- `packages/core/test/unit/schema/registry.spec.ts`: the two-argument `add()` (registration, adoption, precedence over a declared name) and the narrowed element type.
- `packages/docs/guide/schemas.md`, `packages/docs/packages/core.md`, `.agents/architecture.md`.

## Verification

605 core tests pass (601 before this PR's tests), all ten packages' test targets pass, `tsc --noEmit` clean for both `tsconfig.build.json` and the test tsconfig for the touched specs, eslint clean on the changed files, docs site builds, and the emitted `index.d.mts` carries both `add` overloads, `getAll(): RegisteredSchema<any, CONTEXT>[]` and the exported `RegisteredSchema`.

Additive plus one behavior fix; still a minor across the linked release group.

One note on the merge: #907 landed as a squash commit, so merging `master` back in replayed its `registry.spec.ts` additions as an insertion independent of the ones already on this branch, and git auto-merged both. That left two identical `describe('getAll')` blocks, which of course still passed. Removed in `585c404d`.


tada5hi:
Closing for now: deferring to a v3 refactor, where `Schema.name` can be narrowed at the type level directly instead of being routed around at the registry boundary.

Two findings from this branch worth keeping for that work:

1. **Renaming a schema leaves its sub-schemas behind.** `Schema` stamps its name onto the five sub-schemas in `extendSchemasOptions()`, which runs only in the constructor, while the `name` setter inherited from `BaseSchema` is public. So `schema.name = 'after'` leaves `schema.fields.name === 'before'`, and sub-schemas reach back into the registry by the name they carry (`ResolutionScope.for()` at `resolver/module.ts:251`, `resolveBase()` at `:637`), meaning a stale name resolves to the schema this one used to be. This is a live bug on `master` today, independent of the rest of this PR.

2. **`add(name, schema)` plus a narrowed element type** removes the `as string` that name-keyed consumers of `getAll()` currently pay for a case `add()` already refuses. Worth folding into the v3 naming/typing pass rather than shipping as a 2.x patch on top.

The branch `registry-name-invariant` is left in place.

Issue #909, opened 2026-08-12
Refactor: remove the deprecated singular sort vocabulary and the 2.x compatibility machinery


tada5hi:
PR #906 made `sorts` the canonical spelling of the sort parameter on every developer-authored surface and kept every singular name working as a deprecated alias, so that release could stay a clean minor. This issue tracks removing all of it.

## Why this should be near-term, not a long tail

Measured on 2026-08-12, the week ending 2026-08-09 (so pre-GA, beta-era installs):

| Package | Weekly downloads |
|---|---|
| `@rapiq/core` (v2, scoped) | 1,740 |
| `rapiq` (v1, single package) | 173,361 |

v2 adoption is at or below roughly 1 percent of the v1 base, and a chunk of that is plausibly CI. The audience that matters has not migrated yet. Every week the aliases stay is a week they serve nobody, and a v1 user who migrates straight to 3.0 never sees the transitional shape at all. Once a migration guide exists it should point v1 users at 3.0 directly.

## Remove: the deprecated input key on every surface

- `QueryBuildInput.sort` (`packages/core/src/build/types.ts`)
- `SchemaOptionsNormalized.sort` (`packages/core/src/schema/types.ts`)
- `Schema.sort` property (`packages/core/src/schema/module.ts`)
- `ParseQueryOptions.sort` skip flag (`packages/core/src/parser/types.ts`)
- the `sort` fallback in `BaseQueryParser.readParameter` / `skipParameter` (`packages/core/src/parser/query.ts`)
- `EntitySchemaOptions.sort` (`packages/adapter-typeorm/src/schema/types.ts`)
- `ModelSchemaOptions.sort` (`packages/adapter-prisma/src/schema/types.ts`)

## Remove: `Parameter.SORT`

`packages/core/src/constants.ts`. Its value `'sort'` is deliberately unchanged in 2.x, so removal is what actually breaks. Note the URL wire name is a separate constant (`URLParameter.SORT`) and **stays `sort` forever**: it is the JSON:API spelling and is not part of this cleanup.

## Remove: the deprecated aliases

**@rapiq/core**
`SortSchema`, `defineSortSchema`, `SortOptions`, `SortOptionDefault`, `SortSchemaDescription` (`schema/parameter/sort/deprecated.ts`, the whole file goes), `SortParseError`, `SortParseOptions`, `parseSort()` / `parseSortAsync()`, `applySortIndexPolicy`, `buildSortDefaults`

**@rapiq/parser-simple**
[...]

PR #912, opened 2026-08-13
feat: call-time throwOnFailure override, mod as a dialect capability, structured feature tag

tada5hi:
A post-GA consistency audit swept the public API across four dimensions (option shapes, error model, method signatures, cross-adapter behaviour) to decide what should be batched into the v3 major that #909 opens. The verdict was that **the API is broadly consistent and nothing found forces a major**: four findings survived adversarial verification out of eight, and all four are fixable additively. Those four, plus three low-severity items, are this PR. The breaking candidates stay deferred to #909.

Everything here is additive: no public symbol renamed or removed, no required argument added, no thrown class changed, no existing error message changed. Ships as a minor.

### `throwOnFailure` works as a call-time override again (A1)

Every per-parameter parse-option type declares `throwOnFailure` and each sub-parser honours it as an override. The two shared plumbing types did not, and `prepareQueryContext` rebuilt the parameter options copying only `schema`, `strict` and `context`, so the override was silently unavailable at `parser.parse()` and `URLCodec.decode()`: the entry points most applications actually use.

This is a gap against the recorded design, not only an inconsistency. `.agents/migration-notes.md` states the intended policy as `parse-option ?? schema ?? false`, with the override inherited into relation recursion, exactly as `strict` behaves today.

```typescript
// now works at the whole-query and codec level, as it always did per parameter
const query = parser.parse(input, { schema: 'user', throwOnFailure: true });
const decoded = codec.decode(req.query, { schema: 'user', throwOnFailure: true });
```

Also threaded into the relation validations and the index policies, which previously saw only the schema setting.

### `mod` is a dialect capability, and mssql works (A2)

`mod` was rendered unconditionally as `mod(field, x) = y`. SQL Server has no `MOD()` function, only the `%` operator, so an mssql consumer got an opaque database syntax error rather than a typed refusal. No test covered it: `mod` was only exercised under the pg preset.

`DialectOptions` gains an optional `mod` slot. pg, mysql, sqlite and oracle keep today's rendering byte for byte; mssql implements it with `%`, since unlike `regexp` there is no real capability gap on SQL Server.

Worth noting for anyone tempted to just change the hardcoded spelling: no single form works everywhere. Oracle has `MOD()` but not `%`, SQL Server has `%` but not `MOD()`.

### The refused capability is readable without parsing prose (A6)

`AdapterError` carries the feature tag as a structured `feature` property alongside `code`, so a consumer can build a capability matrix without string-matching the message. Message text is byte-identical, since it is currently the only place the tag exists.

```typescript
catch (e) {
    if (e instanceof AdapterError && e.feature === 'filters:mod') { /* ... */ }
}
```

### Smaller items

- **A4**: `FiltersSchemaDescription` gains `caseSensitive`. It is a static declaration sitting beside `allowed`, so its absence from the description contradicted `describe()`'s own documented contract of being the static upper bound.
- **A5**: three raw `SyntaxError` throws replaced with typed errors, so no path in the codebase produces a non-`BaseError` exception. Each catch site was read rather than assumed.
- **A7**: sorting by a to-many relation path now raises the same typed refusal in prisma that drizzle already raises, instead of deferring to an opaque Prisma error at a later layer.
- **A3**: docs fix. `guide/errors.md` filed "regex on a dialect without regex support" under `OPERATOR_UNSUPPORTED`; every backend adapter throws `FEATURE_UNSUPPORTED` for it, and the package pages already said so.

## One regression, caught and fixed in-branch

Worth calling out because the shape is instructive. A2 initially made **every `mod` filter throw on every dialect**, including the four it promised to leave untouched: the top-level `Adapter` class hand-picks option fields when constructing its internal `FiltersAdapter` and did not forward the new slot, so the capability gate never saw it.

The suite stayed green at 119/119 the whole time, because `test/unit/interpreters/mod.spec.ts` constructs `FiltersAdapter` directly and bypasses the public class. The coverage was shaped so it could not observe the bug.

Fixed structurally rather than minimally: the constructor forwards the whole options object instead of re-listing members, so a future dialect option cannot repeat this. The new regression test runs through the public `Adapter` across all five presets and was confirmed red before the fix and green after.

## Verification

`nx run-many -t build` 11/11, `nx run-many -t test` 10/10 (2228 tests, adapter-sql 119 to 124), docs site builds, eslint clean per touched package. `test:db` not run.

Spec: `.agents/plans/032-api-consistency-audit.md` (local).

## Summary by CodeRabbit

* **New Features**
  * Added modulo (`mod`) filtering support across SQL databases, with dialect-specific syntax and clear errors when unsupported.
  * Added per-request `throwOnFailure` controls for parsing and decoding, including relation validation and filter policies.
  * Schema descriptions now report normalized `caseSensitive` settings.

* **Bug Fixes**
  * Prisma now rejects unsupported sorting through to-many relations with a typed error; to-one relation sorting remains supported.
  * Invalid filter values now return consistent typed parsing errors.
  * Adapter errors now identify the unsupported feature when applicable.


tada5hi:
@coderabbitai pause

